PR #3, opened 2016-10-05
Apply opacity and rotation to icon style

chrismayer:
This applies possible opacity and rotation values defined in the Mapbox style to the ``ol.style.Icon`` object.

The assignment is done outside the caching object (``iconImageCache``) by purpose to ensure an icon can be reused with different opacity and rotation values.

### DIFF
--- a/index.js
+++ b/index.js
@@ -204,6 +204,10 @@ function colorWithOpacity(color, opacity) {
   return color;
 }
 
+function deg2rad(degrees) {
+  return degrees * Math.PI / 180;
+}
+
 var spriteRegEx = /^(.*)(\?access_token=.*)$/;
 
 function toSpriteUrl(url, extension) {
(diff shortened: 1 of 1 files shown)


ahocevar:
To clarify my above review comments: You'll want to add both properties to `functions.interpolated`, and then call `paint['icon-opacity'](zoom)` instead of just assigning `paint['icon-opacity']`

chrismayer:
I feared that it isn't that easy 😉 

I had a look at the Mapbox GL Style spec. It seems that [icon-opacity](https://www.mapbox.com/mapbox-gl-style-spec/#paint-icon-opacity) and [icon-rotate](https://www.mapbox.com/mapbox-gl-style-spec/#layout-symbol-icon-rotate) can only be numbers.  Am I missing anything regarding the "ramping functions"?

chrismayer:
> Am I missing anything regarding the "ramping functions"?

Do you mean what is called "Zoom functions" [here](https://www.mapbox.com/mapbox-gl-style-spec/#types-function)?

ahocevar:
Yes. See my clarification https://github.com/boundlessgeo/ol-mapbox-gl-style/pull/3#issuecomment-251643293 - it is actually very easy.

chrismayer:
I just added another commit implementing the support for the ramping functions according to your suggestions in https://github.com/boundlessgeo/ol-mapbox-gl-style/pull/3#issuecomment-251643293. 

Would you have another look, please? Thanks!

chrismayer:
> OpenLayers rotation is expected in radians

Right @ahocevar, good catch! I just changed this. Thanks for your multiple reviews.

PR #7, opened 2016-12-09
Added support for circle styles

theduckylittle:
There was no support for doing basic 'Points as Circles' rendering.  This adds support for the Mapbox GL styles that allow the user to configure those.

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,6 @@ var defaults = {
   'icon-opacity': 1,
   'icon-rotate': 0,
   'icon-size': 1,
-  'circle-radius' : 5,
   'circle-color' : '#000000',
   'circle-stroke-color' : '#000000'
 };
--- a/index.js
+++ b/index.js
@@ -36,7 +36,10 @@ var functions = {
     'text-color',
     'text-field',
     'text-font',
-    'text-halo-color'
+    'text-halo-color',
+    'circle-radius',
+    'circle-color',
+    'circle-stroke-color'
   ]
 };
 
@@ -57,7 +60,10 @@ var defaults = {
   'text-size': 16,
   'icon-opacity': 1,
   'icon-rotate': 0,
-  'icon-size': 1
+  'icon-size': 1,
+  'circle-radius' : 5,
+  'circle-color' : '#000000',
+  'circle-stroke-color' : '#000000'
 };
 
 function applyDefaults(properties) {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -492,6 +498,25 @@ function getStyleFunction(glStyle, source, resolutions, onChange) {
           }
         }
 
+        if (type == 'Point' && 'circle-radius' in paint) {
+          style = new ol.style.Style({
```
ahocevar: This will create a new `ol.style.Style` instance for every rendered feature, which is currently very inefficient in OpenLayers. Better to cache the style.

theduckylittle:
No problem. Let me know if this makes sense: https://github.com/boundlessgeo/ol-mapbox-style/pull/7/commits/6f5a54f5d1d5bce0f5e86ff96e42a04d528ef471

I reused the existing icon caching object but generate a hash based on the radius, stroke, and fill.

index.js
```diff
@@ -492,6 +498,31 @@ function getStyleFunction(glStyle, source, resolutions, onChange) {
           }
         }
 
+        if (type == 'Point' && 'circle-radius' in paint) {
+          ++stylesLength;
+          var cache_key = paint['circle-radius'] + '.' +
+            paint['circle-stroke-color'] + '.' +
+            paint['circle-color'];
+          style = iconImageCache[cache_key];
+          if(!style) {
```
ahocevar: This could just be an `else` block after the `if (style)` block below.

index.js
```diff
@@ -36,7 +36,10 @@ var functions = {
     'text-color',
     'text-field',
     'text-font',
-    'text-halo-color'
+    'text-halo-color',
+    'circle-radius',
```
ahocevar: `circle-radius` is interpolated, not piecewise-constant.

index.js
```diff
@@ -492,6 +498,31 @@ function getStyleFunction(glStyle, source, resolutions, onChange) {
           }
         }
 
+        if (type == 'Point' && 'circle-radius' in paint) {
+          ++stylesLength;
+          var cache_key = paint['circle-radius'] + '.' +
```
ahocevar: `circle-radius`, `circle-stroke-color` and `circle-color` are functions, so you have to use e.g. `paint['circle-radius'](zoom)` for the cache key as well.

index.js
theduckylittle: I just removed the if block.  style should *always* evaluate to true.  It will either come from the cache-object or be created.

index.js
theduckylittle: Doh! I even know that and missed it. Fixed.

index.js
theduckylittle: I shifted it to the other list.

theduckylittle:
All changes fixed up.  Some of those should've been complete obvious, thanks for the patience!

index.js
```diff
@@ -57,7 +60,10 @@ var defaults = {
   'text-size': 16,
   'icon-opacity': 1,
   'icon-rotate': 0,
-  'icon-size': 1
+  'icon-size': 1,
+  'circle-radius' : 5,
```
ahocevar: You have to remove `circle-radius` from the defaults. Otherwise all points will be rendered with a circle, even if no `circle-radius` is in the style object.

theduckylittle: Quite right. Fixed!

ahocevar:
Awesome, thanks a lot @theduckylittle.

PR #13, opened 2017-01-16
bugfix(point): fill should use ol.style.Fill

### LINKED ISSUE #12: point's fill should use ol.style.Fill
This is a quite interesting bug.

As long as I've used `ol-debug.js` it worked just fine. However, using a custom build `ol.js`, it stopped working (does not even render the layer with the basic OL styles), throwing mysterious errors to the console.

I managed to get it up and running again with this patch, but I'd like to hear what you think.

aberenyi:
Fixes #12


aberenyi:
No probs mate

PR #14, opened 2017-01-18
feat(spriteCache): cache the sprites that are downloaded already

aberenyi:
I think it could be quite useful to cache the downloaded sprites if one is about to re-apply styles to (a subset of) layers (for example to keep the number of styles low in Mapbox Studio).

I couldn't extensively test this though - could you please give it a go?

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,8 @@ var glfun = require('mapbox-gl-style-spec/lib/function');
 var mb2css = require('mapbox-to-css-font');
 var FontFaceObserver = require('fontfaceobserver');
 
+var spriteCache = {};
+
 var functions = {
   interpolated: [
     'line-miter-limit',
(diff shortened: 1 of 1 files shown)


ahocevar:
This will not work. `spriteData` is added to the cache before it is loaded. And when that is fixed, things will still not work correctly when `getStyleFunction()` is called again while the data or the image is still loading. So it requires more work to code that up properly. But I wonder if it's worth it. Both the image and the sprite data should be in the browser cache after the first call, and you should get a `304` status code when they are requested again.

aberenyi:
Fair enough, I should have conducted more thorough testing. However, I think this is something worth considering, definitely in my use case. 

The changes that you have on your mind re v2.x sound really great - looking forward to it.

PR #35, opened 2017-11-18
add example with OpenMapTiles style and tile hosting

albanm:
I think supporting OpenMapTiles style would be great. This example works quite well on some aspects, but there are a few bugs.

Labels are not rendered properly, and some zones become black when zooming towards them.

To see the same style fully rendered you can see here : https://www.tilehosting.com/

### DIFF
--- a/example/openmaptiles.html
+++ b/example/openmaptiles.html
@@ -1,0 +1,27 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <meta charset="UTF-8">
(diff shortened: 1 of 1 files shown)


ahocevar:
Label rendering and the black zones should be fixed with the latest release. When you are able to update the pull request to the latest OpenLayers version (v4.6.2), and add the example to the index (example/index.html), I'll gladly merge it. Thanks for the contribution!

fredj:
An API key is now required to get access to the tiles. I reworked the example to add a prompt for the key (same as the mapbox example). See #151

PR #38, opened 2017-11-29
Do not fail on setTarget(null)

bartvde:
In our React component we setTarget(null) on unmount of the component, however this causes an error in the background logic of ol-mapbox-style. This change protects the code against having no map target anymore.


bartvde:
thanks for the review @ahocevar

PR #44, opened 2018-01-29
Respect existing map view

petrsloup:
This fixes a rare case. When using `olms.apply` with existing Map object with a style that does not have `center` and `zoom`, the map view is incorrectly reset.

### DIFF
--- a/index.js
+++ b/index.js
@@ -232,16 +232,15 @@ function getSourceIdByRef(layers, ref) {
   return sourceId;
 }
 
-var hasView = true;
 function processStyle(glStyle, map, baseUrl, host, path, accessToken) {
   var view = map.getView();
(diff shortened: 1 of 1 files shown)


ahocevar:
This changes behavior in another rare case, i.e. when `apply()` is called with an existing map that does not have an initialized view.

I think in addition to your change, `hasView` should be set to `true` only under a more restrictive condition. I added 236e095 which does what I have in mind. Does that look ok to you @petrsloup?

petrsloup:
Oh, I haven't thought about the other case.
Your fix looks like a proper solution. It also ensures that the view is always set, which is nice.

ahocevar:
Thanks for the feedback and the contribution!

PR #55, opened 2018-03-20
Convert to Webpack / Babel / Jest

theduckylittle:
Changes the build and test system to use webpack, babel, and jest.


ahocevar:
Note: this pull request also removes support for a standalone build. Users should now use this together with [ol](https://npmjs.com/package/ol) only.

PR #57, opened 2018-03-20
Set the sonarqube language

bartvde:
Seemed SonarQube was doing a Java analysis and failing, trying to set it javascript explicitly to see if that works


bartvde:
CI change only

PR #61, opened 2018-03-20
Do not define function in for loop

bartvde:
selection_552

Also added a pretest task so linting will always get run


ahocevar:
The `tileLoadFunction` is not executed in the scope of the source, so `this` won't work. Is there a way to tell sonarqube that the function creation inside the for-loop was intentional?

bartvde:
I'm going to close this one, instead I will mark the issue as won't fix in the sonar interface

PR #68, opened 2018-03-22
Provide a full standalone build again

ahocevar:


ahocevar:
I need to provide a better dev server for the examples.

ahocevar:
@theduckylittle Thanks for looking into the examples build.

PR #83, opened 2018-05-22
Add support for fill-pattern

songyumeng:
For https://github.com/boundlessgeo/ol-mapbox-style/issues/82

image

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -302,7 +302,7 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
   const textColor = new Fill();
 
   const iconImageCache = {};
-
+  const patternCache = {};
   const styles = [];
 
   const styleFunction = function(feature, resolution) {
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks! This will even look nicer with openlayers/openlayers#8210.

songyumeng:
It's cool.I found same problems about [openlayers/openlayers#8210](https://github.com/openlayers/openlayers/pull/8210) .
I modify the ol.render.canvas.Replay.prototype.applyFill method in my project like this :
ol.render.canvas.Replay.prototype.applyFill = function (state, geometry) {
        var fillStyle = state.fillStyle;
        var fillInstruction = [ol.render.canvas.Instruction.SET_FILL_STYLE, fillStyle];
        if (typeof fillStyle !== 'string') {
            var viewExtent = this.map.getView().getProjection().getExtent();
            fillInstruction.push([viewExtent[0], viewExtent[3]]);
        }
        this.instructions.push(fillInstruction);
 };
But this function must define after the ‘map‘ object  is defined,and it is not a good solution.
Your PR for openlayers  is very helpful to me.Thanks again.

PR #84, opened 2018-06-05
Always use lowercase imports as a standard to avoid duplicate modules

craigsnyders:
Switch all Openlayers imports to be lowercase to ensure bundlers (e.g. Webpack4 in our case) can tree-shake and JS modules are not duplicated.

This was a problem for us because we use lowercase imports throughout our codebase and even some external modules outside of our control (e.g. https://github.com/Viglino/ol-ext/blob/master/src/interaction/Transform.js).

An example of the webpack errors we had:
```
errors @ webpack-dev-server.js:1
webpack-dev-server.js:1 ./node_modules/ol-mapbox-style/stylefunction.js
Module not found: Error: Can't resolve 'ol/geom/Point' in '/home/csnyders/wrq/webapp2/node_modules/ol-mapbox-style'
errors @ webpack-dev-server.js:1
webpack-dev-server.js:1 ./node_modules/ol-mapbox-style/stylefunction.js
Module not found: Error: Can't resolve 'ol/style/Circle' in '/home/csnyders/wrq/webapp2/node_modules/ol-mapbox-style'
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,17 +10,17 @@ import applyStyleFunction from './stylefunction';
 import googleFonts from 'webfont-matcher/lib/fonts/google';
 import {fromLonLat} from 'ol/proj';
 import {createXYZ} from 'ol/tilegrid';
-import Map from 'ol/Map';
-import GeoJSON from 'ol/format/GeoJSON';
-import MVT from 'ol/format/MVT';
-import {unByKey} from 'ol/Observable';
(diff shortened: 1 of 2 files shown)


ahocevar:
Note that master depends on OpenLayers 5.x, which has CamelCase file names for default exports. If you want to use ol-mapbox-style with OpenLayers 4.x, please use v2.x of this library.

PR #88, opened 2018-06-14
Run tests in puppeteer

ahocevar:
With this change, we no longer require the canvas package, which is a
hassle to set up on some platforms.


ahocevar:
@bartvde I'd really like to run the tests in headless Chrome with Puppeteer, to avoid issues for users with the canvas package. But unfortunately Puppeteer does not work out of the box in Jenkins/Docker (see https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md). Would you be able/willing to help make this work with our CI?

bartvde:
@ahocevar sure

bartvde:
@ahocevar although this will likely require updating the docker Image? I have not touched that but can reach out hopefully to people who have

bartvde:
@BerryDaniel do you have any thoughts on this?

ahocevar:
@bartvde Thanks in advance. I think being able to use Puppeteer can also benefit our other projects.

ahocevar:
@BerryDaniel @bartvde Any news on this? It would really make this library much more dev friendly if we could get this in.

ahocevar:
@BerryDaniel Looks like the Docker issue that prevents Puppeteer from launching could be easily fixed. See https://techoverflow.net/2018/06/05/how-to-fix-puppetteer-error-while-loading-shared-libraries-libx11-xcb-so-1-cannot-open-shared-object-file-no-such-file-or-directory/.

PR #92, opened 2018-08-01
Change imports for mapbox-style-spec

frankrowe:
Currently, this module imports ES6 code directly from the mapbox-style-spec module. This makes it difficult to use in projects that do not transpile the node_modules directory.

This PR imports from the built ES5 code in the mapbox module.

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ import TileJSON from 'ol/source/TileJSON';
 import VectorSource from 'ol/source/Vector';
 import VectorTileSource from 'ol/source/VectorTile';
 import XYZ from 'ol/source/XYZ';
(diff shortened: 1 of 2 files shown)


ahocevar:
There is a huge trade-off between build size and convenience here. When importing from `'@mapbox/mapbox-gl-style-spec'`, we import a lot of stuff we do not need, which tree-shaking also cannot get rid of. In terms of build size, we are talking about 115 kB larger bundles, just by making such a change.

ahocevar:
Actually after removing the babel-loader, bundle size increases only by 10 kB instead of 115 kB. I have created #94 for further discussion.

PR #94, opened 2018-08-02
Use transpiled imports from mapbox-gl-style-spec

ahocevar:
This pull request replaces #92.

The downside is that all bundles created from `ol-mapbox-style` will be larger by 10 kB than with the source imports we previously used.

For `create-react-app` users, we should consider making that work with ES modules, which it maybe already does in the latest alpha version. See facebook/create-react-app#3889.

That said, I'm not sure if we want to merge this. @bartvde @frankrowe what do you think?

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,7 @@ import TileJSON from 'ol/source/TileJSON';
 import VectorSource from 'ol/source/Vector';
 import VectorTileSource from 'ol/source/VectorTile';
 import XYZ from 'ol/source/XYZ';
-import Color from '@mapbox/mapbox-gl-style-spec/util/color';
+import {Color} from '@mapbox/mapbox-gl-style-spec';
 
 var fontFamilyRegEx = /font-family: ?([^;]*);/;
 var stripQuotesRegEx = /("|')/g;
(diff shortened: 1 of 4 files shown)


theduckylittle:
I've tried this with CRA 2.0 this morning and I don't think it solves the problem directly.  Or at least the way Webpack is configured is not doing transpiling the deeper modules.

I'm getting the same syntax errors from that I was getting with CRA 1.x.

ahocevar:
Can you point me to an example project so I can reproduce the issue and investigate? Thanks!

bartvde:
hey @ahocevar basically the instructions here: https://www.npmjs.com/package/@boundlessgeo/sdk

just make sure to leave out the version numbers for ol-mapbox-style and sdk

tschaub:
I just came across this same issue trying to use `ol-mapbox-style` with Parcel.  It fails to parse the flow syntax in `@mapbox/mapbox-gl-style-spec/util/color.js`.

tschaub:
Installing `ol-mapbox-style@3.0.0-beta.7` fixed the issue for me.

bartvde:
see also: https://github.com/mapbox/mapbox-gl-js/issues/7105

bartvde:
@ahocevar any objections to merging this one?

theduckylittle:
At this point it's 10kb vs a lot of broken apps...

ahocevar:
Let‘s do this.

PR #107, opened 2018-10-31
updated @mapbox/mapbox-gl-style-spec 

fleur:
Hi.  I didn't do this earlier because it broke something, and now I've gone back and figured it out.  The problem is that `@mapbox/mapbox-gl-style-spec`  doesn't handle missing property defaults gracefully any more.

It used to be that compiling an expression that had `undefined` for a default would not fail, but it does now.  Not all elements in the stylespec reference have defaults to fall back on, which is what `ol-mapbox-style` was doing if there were no properties at all.  Now even existing properties passed w/o a default will error out, hence the extra if.

I'm pretty sure this is all there is to it, but my domain knowledge is lacking, so definitely sanity check me.  Tests pass, examples work.

Also added nodemon.

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -14,7 +14,8 @@ import Point from 'ol/geom/Point';
 import derefLayers from '@mapbox/mapbox-gl-style-spec/deref';
 import spec from '@mapbox/mapbox-gl-style-spec/reference/latest';
 import {
-  expression, Color,
+  expression,
+  Color,
   function as fn,
   featureFilter as createFilter
 } from '@mapbox/mapbox-gl-style-spec';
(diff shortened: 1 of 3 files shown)


stylefunction.js
```diff
@@ -66,7 +67,9 @@ export function getValue(layer, layoutOrPaint, property, zoom, feature) {
     let value = (layer[layoutOrPaint] || emptyObj)[property];
     const propertySpec = spec[`${layoutOrPaint}_${layer.type}`][property];
     if (value === undefined) {
-      value = propertySpec.default;
+      value = (propertySpec.default !== undefined) ? propertySpec.default : 'none';
```
ahocevar: Leaving the value `undefined` here should not break anything, because a literal value will never run through a function conversion or expression evaluation.

stylefunction.js
```diff
@@ -66,7 +67,9 @@ export function getValue(layer, layoutOrPaint, property, zoom, feature) {
...
+    } else if ((typeof value == 'object') && !('default' in value)) {
+      value.default = (propertySpec.default !== undefined) ? propertySpec.default : '';
```
ahocevar: While this fixes the `inline-geojson` example, it might break in cases where the expected value is not a string. A better default here would be `null`, which is handled by the expression engine, but `icon-image` cannot be null, There is an open issue for this: mapbox/mapbox-gl-js#6935.

fleur:
Ah, okay.  I didn't realize things were dependent on mapbox gl issues.  Not a very useful PR then.  I'll delete it.

PR #110, opened 2018-11-03
Update mapbox-gl-style-spec to 13.3.0

Romaxx:


ahocevar:
Thanks @Romaxx. Updating this dependency requires a few changes, so I took the liberty to add a commit that fixes things.

PR #116, opened 2018-11-13
colorWithOpacity tests and return consistency fix

fleur:
Hi.  Added a few more tests.  I didn't include these in my last PR because the way the cacheing and multiplying the alpha is done initially confused me.

Also, the way `colorWithOpacity()` was written, it was an identity function if no opacity is passed in, meaning it returned a `Color`, but if opacity was provided, an array is returned.  Strictly speaking it doesn't affect functionality, but it took me awhile to convince myself it didn't matter, so for code maintenance purposes, it's a benefit.  I hope you'll agree!


ahocevar:
This pull request is outdated - the `colorWithOpacity()` function now is completely different.

PR #130, opened 2019-02-15
Fix how zoom level is interpreted from a Mapbox style

pakb:
As Mapbox count zoomlevel differently than OpenLayers, there was a misinterpretation and OL was showing layers 1 zoom level too early, this fix this issue.

You can see this difference in zoomlevel interpretation on this codepen (zoom by the Alps around zoom level 8 on mapbox to see peaks appearing on OL when the style says it should not appear before zoom level 9) : https://codepen.io/pakb/pen/NoLJYB

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -290,6 +290,11 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
     if (zoom == -1) {
       zoom = getZoomForResolution(resolution, resolutions);
     }
+    if (zoom > 0) {
+      // mapbox zoom level is counted from 0 to X, OL zoom level is counted from 1 to X
+      // so here we substract 1 to adjust to zoom level described in Mapbox Style
+      zoom -= 1;
+    }
     const type = types[feature.getGeometry().getType()];
     const f = {
       properties: properties,


ahocevar:
ol-mapbox-style uses the correct set of resolutions (i.e. the same that mapbox-gl uses):
https://github.com/openlayers/ol-mapbox-style/blob/master/stylefunction.js#L182-L187, so your change would actually introduce a shift in zoom levels by -1, instead of fixing it.

What you are seeing is an implementation difference between OpenLayers and mapbox-gl. OpenLayers evaluates styles only when the integer zoom level of the used tiles changes, which is between zoom levels (e.g. the switch between 8 and 9 is around 8.4. At integer zoom levels, you'll get the same result in OpenLayers and mapbox-gl. Only between integer zoom levels it is different. I have created an updated version of your codepen (https://codepen.io/ahocevar/pen/RveqGa) which shows the integer mapbox zoom OpenLayers uses for evaluating styles.

pakb:
Ok, that's why the threshold where things started to pop was at 8.4-ish in my case.
Wouldn't it be better to floor the int instead of rounding it? This way even if zoom level is 8.9, the style would be applied with a mapbox zoom level of 8? Otherwise the style is interpreted differently.

ahocevar:
You can play with this by changing the OpenLayers renderer's `zDirection` (not part of the API). I wouldn't make it default though, because it also means more undersampling for the pre-rendered tile images.
```js
olms('map', style).then(function(map) {
  map.getLayers().forEach(function(layer) {
    layer.getRenderer().zDirection = 1;
  });
});
```
I have updated your codepen to show this: https://codepen.io/ahocevar/pen/RveqGa.

pakb:
Ok much better, thank you.
I still don't understand why OL shows more peaks than Mapbox at zoom level 9 but at least they seem to appear at the same zoom level with your config.
Is the decluttering also configurable, with a bigger "safe zone" between items?

ahocevar:
@pakb The style for the peaks has a `text-padding` of `50` configured, and ol-mapbox-style does not support the `text-padding` property. We'd accept a pull request for that.

pakb:
Ok I'll give a try, I'm closing this one as this can be solved by configuration
Although this configuration could be a nice adding to the documentation as this might be an issue for a typical use case (straight import of a mapbox style JSON)

ahocevar:
@pakb:
> Although this configuration could be a nice adding to the documentation as this might be an issue for a typical use case (straight import of a mapbox style JSON)

I don't think this is an option that you would want to use in production. Look at your map at the street level and see how blurry it can get (note this is the final image, not during loading):
untitled

pakb:
I’ll have to find out a solution a way or another, I could disable this settings past a certain zoom level for instance

PR #131, opened 2019-02-15
adding olms-debug target

boecklic:
For development of "legacy" (e.g. non modular) JS apps it would be useful to have an aggregated `olms-debug.js` build target as well. I propose a solution with this PR, but since I'm not very familiar with webpack yet it might not be the cleanest solution...

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "start": "webpack-dev-server --config ./webpack.config.js",
     "prepare": "npm run doc && npm run build",
     "build": "webpack-cli --mode=production --config ./webpack.config.olms.js && webpack-cli --mode=production",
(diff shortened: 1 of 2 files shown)


ahocevar:
@boecklic You can easily create a debug build with a single command:

    webpack-cli --mode=development --config ./webpack.config.olms.js --output-filename olms-debug.js

This will create `olms-debug.js` and `olms-debug.js.map` in the `dist/` folder.

boecklic:
Ok thanks, so I will close this PR.

ahocevar:
If you're looking for this because you want to do quick prototyping in a live code editor, I'd recommend https://codesandbox.io/. That one can use npm modules (which e.g. CodePen cannot).

PR #141, opened 2019-02-19
Add text padding support

pakb:
Would fix part of https://github.com/openlayers/ol-mapbox-style/issues/137

Text-padding should be a number, as defined in the [Mapbox spec](https://docs.mapbox.com/mapbox-gl-js/style-spec#layout-symbol-text-padding) so it is converted to an array as wanted in [OL spec](https://openlayers.org/en/latest/apidoc/module-ol_style_Text-Text.html)

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -637,6 +637,10 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
           } else {
             text.setStroke(undefined);
           }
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks @pakb. I added a commit that optimizes things in terms of performance, and avoids garbage creation. Please give a thumbs up if that change works for you.

pakb:
Yep, forgot that it was called in a loop, nice optimization

PR #142, opened 2019-02-19
Add text-letter-spacing support

pakb:
Add support for https://docs.mapbox.com/mapbox-gl-js/style-spec#layout-symbol-text-letter-spacing

will fix https://github.com/openlayers/ol-mapbox-style/issues/137

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -607,8 +607,12 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
             let wrappedLabelWithLetterSpacing = '';
             const wrappedLabelLines = wrappedLabel.split('\n');
             const joinSpaceString = hairSpacePool.slice(0, Math.round(letterSpacing / 0.1));
-            for (let lineIndex = 0; lineIndex < wrappedLabelLines.length; lineIndex += 1) {
-              wrappedLabelWithLetterSpacing += wrappedLabelLines[lineIndex].split('').join(joinSpaceString);
+            const ll = wrappedLabelLines.length;
+            for (let l = 0; l < ll; ++l) {
+              wrappedLabelWithLetterSpacing += wrappedLabelLines[l].split('').join(joinSpaceString);
+              if (l + 1 < ll) {
+                wrappedLabelWithLetterSpacing += '\n'
+              };
             }
             text.setText(wrappedLabelWithLetterSpacing);
           } else {
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -607,12 +607,11 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
             let wrappedLabelWithLetterSpacing = '';
             const wrappedLabelLines = wrappedLabel.split('\n');
             const joinSpaceString = hairSpacePool.slice(0, Math.round(letterSpacing / 0.1));
-            const ll = wrappedLabelLines.length;
-            for (let l = 0; l < ll; ++l) {
+            for (let l = 0, ll = wrappedLabelLines.length; l < ll; ++l) {
+              if (l > 0) {
+                wrappedLabelWithLetterSpacing += '\n';
+              }
               wrappedLabelWithLetterSpacing += wrappedLabelLines[l].split('').join(joinSpaceString);
-              if (l + 1 < ll) {
-                wrappedLabelWithLetterSpacing += '\n'
-              };
             }
             text.setText(wrappedLabelWithLetterSpacing);
           } else {
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -25,6 +25,7 @@ const isFunction = fn.isFunction;
 const convertFunction = fn.convertFunction;
 const isExpression = expression.isExpression;
 const createPropertyExpression = expression.createPropertyExpression;
+const hairSpacePool = Array(256).join('\u200A');
 
 const types = {
   'Point': 1,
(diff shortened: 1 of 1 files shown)


stylefunction.js
```diff
@@ -601,6 +601,19 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
             label = label.toLowerCase();
           }
           const wrappedLabel = type == 2 ? label : wrapText(label, font, getValue(layer, 'layout', 'text-max-width', zoom, f));
+          const letterSpacing = getValue(layer, 'layout', 'text-letter-spacing', zoom, f);
+          if (letterSpacing && letterSpacing > 0.01 /* default is 0.01 */ ) {
```
ahocevar: `letterSpacing` should always be available here, so you can remove `letterSpacing && `.

And we should enter this code path only when `letterSpacing >= 0.05`. See below.

stylefunction.js
```diff
@@ -601,6 +601,19 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
...
+            let wrappedLabelWithLetterSpacing = '';
+            wrappedLabel.split('\n').forEach((line) => {
+              let joinSpaceString = "",
```
ahocevar: Please rebase your pull request on top of master so the CI catches this, and use single quotes.

stylefunction.js
```diff
@@ -601,6 +601,19 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
...
+                  spacingCount;
+              for (spacingCount = 0; spacingCount < letterSpacing; spacingCount += 0.1) {
+                joinSpaceString += String.fromCharCode(8202); // Hair space, about 0.1em per character
```
ahocevar: `'\u200A'` should be faster than `String.fromCharCode(8202)`.

stylefunction.js
```diff
@@ -601,6 +601,19 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
             label = label.toLowerCase();
           }
           const wrappedLabel = type == 2 ? label : wrapText(label, font, getValue(layer, 'layout', 'text-max-width', zoom, f));
+          const letterSpacing = getValue(layer, 'layout', 'text-letter-spacing', zoom, f);
+          if (letterSpacing && letterSpacing > 0.01 /* default is 0.01 */ ) {
+            let wrappedLabelWithLetterSpacing = '';
+            wrappedLabel.split('\n').forEach((line) => {
```
ahocevar: For performance reasons, please use a `for` loop instead of `forEach`.

stylefunction.js
```diff
@@ -601,6 +601,19 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
...
+              let joinSpaceString = "",
+                  spacingCount;
+              for (spacingCount = 0; spacingCount < letterSpacing; spacingCount += 0.1) {
```
ahocevar: No need to do this in a loop. The `joinSpaceString` can be created with
```js
const joinSpaceString = Array(Math.round(letterSpacing / 0.1)).join('\u200A');
```
This way we add one hair space for a `letterSpacing` of `0.5` to `1.499999`, two for `1.5` to `2.499999` and so on.

pakb: Isnt't join also part of the filter map reduce, and thus will be a bigger impact as to do a simple for loop?

ahocevar: You're right regarding performance. And both your and my approach are lame. I've come up with a 3rd way, see https://jsperf.com/pad-strings. So the best is to define a string pool only once that we slice from:
```js
// on the module level
const hairSpacePool = Array(256).join('\u200A');

export function styleFunction(feature, resolution) {
  // ...
  const joinSpaceString = hairSpacePool.slice(0, Math.round(letterSpacing / 0.1));
  // ...
}

**Comment by pakb** [comment]
Should be good, let me know if you think we need to improve something else with this PR

File: stylefunction.js [hunk 6]
```diff
@@ -601,7 +602,18 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
             label = label.toLowerCase();
           }
           const wrappedLabel = type == 2 ? label : wrapText(label, font, getValue(layer, 'layout', 'text-max-width', zoom, f));
-          text.setText(wrappedLabel);
+          const letterSpacing = getValue(layer, 'layout', 'text-letter-spacing', zoom, f);
+          if (letterSpacing >= 0.05) {
+            let wrappedLabelWithLetterSpacing = '';
+            const wrappedLabelLines = wrappedLabel.split('\n');
+            const joinSpaceString = hairSpacePool.slice(0, Math.round(letterSpacing / 0.1));
+            for (let lineIndex = 0; lineIndex < wrappedLabelLines.length; lineIndex += 1) {
```
ahocevar:
```suggestion
            for (let l = 0, ll = wrappedLabelLines.length, l < ll; ++l) {
```

stylefunction.js
```diff
@@ -601,7 +602,18 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
...
+              wrappedLabelWithLetterSpacing += wrappedLabelLines[lineIndex].split('').join(joinSpaceString);
```
ahocevar:
```suggestion
              wrappedLabelWithLetterSpacing += wrappedLabelLines[l].split('').join(joinSpaceString);
```

ahocevar: Oh, and one more thing: you need to re-add the line breaks to wrappedLabelWithLetterSpacing.

pakb: oops, yep forgot that

PR #143, opened 2019-02-20
fix getLayers() for non-mapbox layers

RobertOrthofer:
This prevents non-mapbox layers breaking the code at `getLayers()`. 
Also added a small test with an empty `VectorTileLayer`.


ahocevar:
Thanks for this contribution @RobertOrthofer. While reviewing, I saw that the `indexOf` check does not make sense, because `layer.get('mapbox-source')` returns a string, not an array. Your fix here inspired me to create #144, which replaces this pull request. Can you take a look at #144 please before I merge that?

PR #144, opened 2019-02-20
Fix getSource, getLayer and getLayers

ahocevar:
This pull request is a replacement for #143 and fixes more underlying issues of the `getLayer()`, `getLayers()` and `getSource()` function. The `indexOf` check does not make sense for `getSource()` and `getLayers()` because the `mapbox-source` property of the layer is a string, not an array. And for `getLayer()`, the fix proposed by @RobertOrthofer in #143 needs to be applied.

Thanks @RobertOrthofer for inspiring this change.


RobertOrthofer:
seems good to me, thx!

ahocevar:
I just published v3.9.0, which contains this fix.

PR #146, opened 2019-02-21
add *-keep-upright

GD-Aichi:
I do not know if it is desired, because mapbox support keep-upright only for lines, but I need this.

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -175,10 +175,11 @@ function fromTemplate(text, properties) {
  * is available. Font names are the names used in the Mapbox Style object. If
  * not provided, the font stack will be used as-is. This function can also be
  * used for loading web fonts.
+ * @param {ol.View} [view=undefined] View of the map
  * @return {ol.style.StyleFunction} Style function for use in
  * `ol.layer.Vector` or `ol.layer.VectorTile`.
  */
-export default function(olLayer, glStyle, source, resolutions, spriteData, spriteImageUrl, getFonts) {
+export default function(olLayer, glStyle, source, resolutions, spriteData, spriteImageUrl, getFonts, view) {
   if (!resolutions) {
     resolutions = [];
     for (let res = 78271.51696402048; resolutions.length <= 24; res /= 2) {
@@ -520,7 +521,13 @@ export default function(olLayer, glStyle, source, resolutions, spriteData, sprit
                     });
                   }
                 }
-                iconImg.setRotation(placementAngle + deg2rad(getValue(layer, 'layout', 'icon-rotate', zoom, f)));
+                let rotate = placementAngle + deg2rad(getValue(layer, 'layout', 'icon-rotate', zoom, f));
+                if (iconRotationAlignment == 'map') {
+                  const keepUpright = getValue(layer, 'layout', 'icon-keep-upright', zoom, f);
+                  if (keepUpright && view) {
+                    rotate = flipRotation(view.getRotation(), rotate);
+                  }
+                }
                 iconImg.setOpacity(getValue(layer, 'paint', 'icon-opacity', zoom, f));
                 style.setImage(iconImg);
                 text = style.getText();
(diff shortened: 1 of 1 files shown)


ahocevar:
What is the use case for setting `text-keep-upright` or `icon-keep-upright` on points that align with the map? In mapbox-gl, these settings are meant to keep text and icons readable when they follow lines.

GD-Aichi:
I have a picture at one point, which rotates with the map and as soon as it is upside down, then the picture should turn. I have the same with texts.
```json
{
            "id": "textures",
            "type": "symbol",
            "source": "overlays",
            "source-layer": "textures",
            "minzoom": 1,
            "filter": [
                "has",
                "symbol"
            ],
            "layout": {
                "icon-image": "{symbol}",
                "icon-rotation-alignment": "map",
                "icon-rotate": {
                    "type": "identity",
                    "property": "image_rotation"
                },
                "icon-keep-upright": true,
                "icon-size": {
                    "stops": [
                        [
                            2,
                            0.02946367895652174
                        ],
                        [
                            3,
                            0.05892735791304348
                        ],
                        [
                            4,
                            0.11785471582608696
                        ],
                        [
                            5,
                            0.23570943165217392
                        ],
                        [
                            6,
                            0.47141886330434785
                        ],
                        [
                            7,
                            0.9428377266086957
                        ],
                        [
                            8,
                            1.8856754532173914
                        ]
                    ]
                }
            },
            "paint": {
                "icon-opacity": { "stops": [[1.8, 0], [2, 1]] }
            }
        },
```

ahocevar:
So you want to flip the image rotation, as defined in your `image_rotation` property? I'm hesitant to add features to the library that come from invalid Mapbox Style documents. But you should be able to modify styles as you get them from the style function in a post-processing step. So you could do something like
```js
import olMapboxStyle, {getLayer} from 'ol-mapbox-style';

olMapboxStyle('map', 'my/style.json').then(function(map) {
  const layer = getLayer(map, 'textures');
  const style = layer.getStyle();
  const view = map.getView();
  layer.setStyleFunction(function(feature, resolution) {
    const styles = style(feature, resolution);
    if (feature.get('image_rotation') !== undefined) {}
    for (let i = 0, ii = styles.length; i < ii; ++i) {
      const icon = style.getImage();
      if (icon) {
        icon.setRotation(flipRotation(view, icon.getRotation()));
      }
    }
    return styles;
  });
});
```

GD-Aichi:
It's not exactly what I need. The layer 'textures' is just one example, I need a dynamic variate. I use the "stylefunction" function.

That was my first way, but I need something like getLayerByFeature(feature).
```javascript
stylefunction(vectorLayer, glStyle, 'vector',
  map.getView().getResolutions(), sprite, '...', getFonts);

const styleFn = vectorLayer.getStyle();
vectorLayer.setStyle((feature, resolution) => {
  const style = styleFn(feature, resolution);
  // .. change style
  return style;
});
```

ahocevar:
You already have something like `getLayerByFeature()`. When using vector tiles, each feature has a `layer` property. So you can get the OpenLayers layer instance with
```js
const layer = getLayer(map, feature.get('layer'));
```

GD-Aichi:
Sorry, I did not speak of OL layers, but of MB layers. e.g.
```json
{
    "id": "textures",
    "type": "symbol",
    "source": "overlays",
    "source-layer": "textures",
    "minzoom": 1,
    "maxzoom": 2,
    "layout": {
    }
},
{
    "id": "textures",
    "type": "symbol",
    "source": "overlays",
    "source-layer": "textures",
    "minzoom": 3,
    "layout": {
    }
},
{
    "id": "textures",
    "type": "symbol",
    "source": "overlays",
    "source-layer": "textures",
    "minzoom": 3,
    "filter": [
        "has",
        "symbol"
    ],
    "layout": {
    }
}
```
what is the style layer that is currently being used? 
"getStyleLayerByFeature()" ;)

ahocevar:
Ok, something like this then:
```js
function getStyleLayer(feature) {
  const layers = glStyle.layers;
  for (let i = 0, ii = layers.length; i < ii; ++i) {
    if (layers[i].id === feature.get('layer') {
      return layers[i];
    }
  }
}
```

GD-Aichi:
Yes exactly, but in your example, the function returns the first layer it finds, but I have a minzoom, maxzoom, or some other filter. The function should return exactly the MB layer used for styling.

ahocevar:
The layer returned by the function I posted above is the one that the feature belongs to. If the filter configured with that layer does not pass, then the style will be empty. That should give you all information you need.

Beyond these hints, I cannot provide free consulting here.

Either way: it seems to me that the way you want to use `icon-keep-upright` and `text-keep-upgright` is not what the spec intended. Correct me if I'm wrong. But if you agree, please close this issue.

PR #154, opened 2019-02-28
Remember when fonts have been loaded already

### LINKED ISSUE #153: getFonts is not caching the results
The function is checking if the font is in the `fontFamilies` cache object but never stores anything in that cache: https://github.com/openlayers/ol-mapbox-style/blob/fc28c683f4e687ab188636a405b52a8bc8b3c09f/index.js#L58-L60

Is this cache supposed to store the font name by families? Something like:
```js
{
  "Noto Sans": ["Noto Sans Bold", "Noto Sans Regular"]
}
```

ahocevar:
Fixes #153.

### DIFF
--- a/index.js
+++ b/index.js
@@ -731,4 +731,7 @@ export function getSource(map, sourceId) {
   }
 }
 
-export {finalizeLayer as _finalizeLayer};
+export {
+  finalizeLayer as _finalizeLayer,
+  getFonts as _getFonts
+};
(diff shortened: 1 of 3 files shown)


ahocevar:
I think this can be simplified even more. Looking into it now.

ahocevar:
Some of the code there is a remainder from the times where OpenLayers did not have built-in web font handling.

ahocevar:
@fredj Now the function does what it is supposed to do. If you could add tests to this pull request, it would be great.

We'll also be getting transferred the ownership of the mapbox-to-css-font library. When done, we can add tests there too.

fredj:
Thanks @ahocevar ! I will add tests for the getFonts function

PR #159, opened 2019-03-04
Always add stroke to fills, without a separate style

### LINKED ISSUE #139: Rendering of small features, min-width?
In mapbox, small feature like rivers seem to have a min-width for rendering as they show up on the map, where they are hidden (because too small for rendering?) in OL after loading mapbox style via olms.

Example in this codepen : https://codepen.io/pakb/full/gqqzyo

ahocevar:
Fixes #139.

Performance is improved for polygons with an outline, probably weighing out the performance penalty for polygons without an outline.

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -356,35 +356,24 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
           } else {
             color = colorWithOpacity(getValue(layer, 'paint', 'fill-color', zoom, f), opacity);
             if (color) {
+              if ('fill-outline-color' in paint) {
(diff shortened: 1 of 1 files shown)


fredj:
There's an issue with the buildings in the openmaptiles example:
screenshot from 2019-03-04 11-48-57
The zoom level is 15

The `fill-outline-color` for the buildings is:
```
"fill-outline-color": {
  "stops": [
    [15,"rgba(212, 177, 146, 0)"],
    [16,"rgba(212, 177, 146, 0.5)"]
  ]
}
```

PR #168, opened 2019-03-07
Text-halo : take text size into account

pakb:
OLMS tackle this feature by creating a ol/style/Stroke behind the text, this doesn't mimic CSS text-shadow (as seems to be implemented in Mapbox) because it doesn't count the shadowed text width.
This is an attempt to get closer to the CSS rendering with text-shadow used in Mapbox.

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -636,7 +636,12 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
           const haloColor = colorWithOpacity(getValue(layer, 'paint', 'text-halo-color', zoom, f), opacity);
           if (haloColor) {
             textHalo.setColor(haloColor);
-            textHalo.setWidth(getValue(layer, 'paint', 'text-halo-width', zoom, f));
+            // spec here : https://docs.mapbox.com/mapbox-gl-js/style-spec/#paint-symbol-text-halo-width
+            // stroke must be doubled because it is applied around the center of the text outline
+            const textHaloWidth = getValue(layer, 'paint', 'text-halo-width', zoom, f) * 2;
+            // 1/4 of text size (spec) x 2
+            const halfTextSize = 0.5 * textSize;
+            textHalo.setWidth(textHaloWidth <= halfTextSize ? textHaloWidth : halfTextSize);
             text.setStroke(textHalo);
           } else {
             text.setStroke(undefined);


stylefunction.js
```diff
@@ -636,7 +636,7 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
           const haloColor = colorWithOpacity(getValue(layer, 'paint', 'text-halo-color', zoom, f), opacity);
           if (haloColor) {
             textHalo.setColor(haloColor);
-            textHalo.setWidth(getValue(layer, 'paint', 'text-halo-width', zoom, f));
+            textHalo.setWidth(getValue(layer, 'paint', 'text-halo-width', zoom, f) + 0.1 * textSize);
```
ahocevar: The stroke is applied around the center of the text outline, so I think
```suggestion
            textHalo.setWidth(getValue(layer, 'paint', 'text-halo-width', zoom, f) + * 2);
```
would bring us closer to what mapbox-gl does.

The [spec](https://docs.mapbox.com/mapbox-gl-js/style-spec/#paint-symbol-text-halo-width) also says that the maximum halo is 1/4 of the text size, which should also be taken into account.

pakb: would a modulo computation be too much performance wise? That could do the trick, something like :
```suggestion
            textHalo.setWidth((getValue(layer, 'paint', 'text-halo-width', zoom, f) * 2) % (0.25 * textSize));
```

pakb:
I don't know if checking our multiplied stroke-width with a quarter of the font size is relevant regarding the spec. Won't it divert the bigger the font is?

ahocevar:
@pakb You'd also have to multiply the `textSize * 0.25` by `2`. As I said, this has nothing to do with css `text-shadow`. [CanvasRenderingContext2D#strokeText](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/strokeText) draws a line with a given width (`strokeStyle` around the outline of the text:

    |-- half width --|-- text outline --|-- half width --|

For 1/4 of the text size, you also have to double that value:

    |-- 1/4 text width --|-- text outline --|-- 1/4 text width --|

pakb:
Ok got it, shadow radius vs stroke width. I’ll update the code and comments

pakb:
Should be good

PR #179, opened 2019-03-13
Load google fonts with correct weight and style

### LINKED ISSUE #177: Load the correct font weight and style from Google Fonts
The `getFonts` function ignores the font weight and style when downloading the font family from google: only the weight 400 is loaded.

For example, when called with `['Noto Sans Bold', 'Noto Sans Regular Italic']`, the style sheet href should be: https://fonts.googleapis.com/css?family=Noto+Sans:700,400i

(700 for `Bold` and 400i for `Regular Italic`)

ahocevar:
Fixes #177.

### DIFF
--- a/index.js
+++ b/index.js
@@ -59,13 +59,15 @@ function getFonts(fonts) {
   if (fontsKey in processedFontFamilies) {
     return fonts;
   }
-  const families = fonts.map(function(font) {
-    return mb2css(font, 1).split(' 1px ')[1].replace(/"/g, '');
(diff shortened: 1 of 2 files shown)


ahocevar:
> For example: getFonts(['Noto Sans']), the href generated is:
https://fonts.googleapis.com/css?family=Noto+Sans:normalnormal

This url is fine though. But I agree it's nicer after openlayers/mapbox-to-css-font#3.

PR #181, opened 2019-03-15
Support text-max-angle

### LINKED ISSUE #175: Support text-max-angle
It seems that `text-max-angle` property is not supported in the converter.
Can be seen here:
https://stackblitz.com/edit/angular-openlayers-mapbox-local-issue-with-labels

According to:
https://github.com/openlayers/openlayers/issues/9309
The relevant variable to set is: `maxAngle` when initializing the `Text` variable.

I'll see if I can create a PR for this.

HarelM:
Resolves #175

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -579,6 +579,7 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
             }
             text.setTextAlign(textAlign);
           } else {
+            text.setMaxAngle(deg2rad(getValue(layer, 'layout', 'text-max-angle', zoom, f)));
             text.setTextAlign();
           }
           let textBaseline = 'middle';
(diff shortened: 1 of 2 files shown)


stylefunction.js
```diff
@@ -612,6 +612,10 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
           if (textPadding !== padding[0]) {
             padding[0] = padding[1] = padding[2] = padding[3] = textPadding;
           }
+          const textMaxAngle = getValue(layer, 'layout', 'text-max-angle', zoom, f);
+          if (textMaxAngle !== undefined && placement === 'line') {
```
ahocevar: textMaxAngle should never be undefined here, so it would be more efficient to only call `getValue()` when `placement` is `'line'`.

HarelM: @ahocevar see commit: d54c15c.
Thanks for the review and fast response!
I noticed that `text-max-angle` can also be set to `line-center` so I have fixed the code accordingly. It looks a lot better now.

HarelM:
Kindly let me know when a new version is available.

ahocevar:
Thanks for the contribution! We're making frequent releases right now. For testing, you can always use a git package url, e.g.

    npm install git://github.com/openlayers/ol-mapbox-style#master

PR #182, opened 2019-03-19
Fixed bug in colorCache

egaoneko:
The bug occur when some features have style with `color : rgb(255, 255, 255)` and `opacity : 1` and other features have style with `color : rgb(255, 255, 255)` and `opacity : 0.5`. They reference same `colorData.color` array. So they make some side effect when features are rendered.

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -104,15 +104,14 @@ function colorWithOpacity(color, opacity) {
     if (color.a === 0 || opacity === 0) {
       return undefined;
     }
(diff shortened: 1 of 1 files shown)


ahocevar:
That was intentional. A color is only used for rendering a single feature, so I wonder where the reference to the same array could be a problem. Can you provide a test case that shows where this causes a problem?

ahocevar:
I see now. This can be a problem for circles and polygons with a fill and outline.

PR #201, opened 2019-04-18
add featurestate

GD-Aichi:


ahocevar:
Thanks for your work on this, @GD-Aichi, and sorry it went unnoticed for so long. I took your work as inspiration for #514, which finally adds support for `feature-state`, in a slightly different way than you did it here.

PR #203, opened 2019-04-25
Account for fixed declutter order in OpenLayers

ahocevar:


GD-Aichi:
Do you work on the fix?

ahocevar:
I'll be merging this pull request as soon as an ol beta with openlayers/openlayers/#9492 is available. Then I'll also be merging the other pull requests that depend on ol@6, and update the peer dependency.

PR #233, opened 2019-10-05
Fix fromTemplate errors by only passing in Strings

eliask:
Objects like `ResolvedImage` can result from `getValue`.
Stringify all return values from `getValue` before passing them to `fromTemplate` to prevent runtime errors.

### DIFF
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -316,7 +316,7 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
           if ('fill-pattern' in paint) {
             const fillIcon = getValue(layer, 'paint', 'fill-pattern', zoom, f);
             if (fillIcon) {
-              const icon = fromTemplate(fillIcon, properties);
+              const icon = fromTemplate(fillIcon.toString(), properties);
               if (spriteImage && spriteData && spriteData[icon]) {
                 ++stylesLength;
                 style = styles[stylesLength];
@@ -413,7 +413,7 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
         if ((type == 1 || type == 2) && 'icon-image' in layout) {
           const iconImage = getValue(layer, 'layout', 'icon-image', zoom, f);
           if (iconImage) {
-            icon = fromTemplate(iconImage, properties);
+            icon = fromTemplate(iconImage.toString(), properties);
             let styleGeom = undefined;
             if (spriteImage && spriteData && spriteData[icon]) {
               const iconRotationAlignment = getValue(layer, 'layout', 'icon-rotation-alignment', zoom, f);
--- a/stylefunction.js
+++ b/stylefunction.js
@@ -316,7 +316,9 @@ export default function(olLayer, glStyle, source, resolutions = defaultResolutio
           if ('fill-pattern' in paint) {
             const fillIcon = getValue(layer, 'paint', 'fill-pattern', zoom, f);
             if (fillIcon) {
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks @eliask. For a bit more context, can you please provide a style that shows the issue, i.e. where `getValue()` returns an object like `ResolvedImage`?

eliask:
Here's what I used:
```
import olms from 'ol-mapbox-style';
olms('map', 'https://tiles.stadiamaps.com/styles/alidade_smooth.json');
```

eliask:
Note: It might be the case that Any non-String values _should be_ treated differently from the template strings in the first place.
It sounds like `ResolvedImage` should never have a template value (I haven't confirmed).

ahocevar:
I'm hoping that anything returned from `getValue()` has a `toString()` function implemented. It should not be necessary to call `toString()`, because that is done implicitly. Are you seeing an exception thrown, or did you just add this to make the type checker happy?

ahocevar:
Maybe it would make sense to change your pull request to change the return statement in `getValue()` to something like
```js
const value = functions[property](zoomObj, feature);
return typeof value === string ? value : value.toString();
```

eliask:
It is definitely an actual error.
Here's a real world example up for at least a little while
https://button-map.netlify.com/ (though it's minified with source map).

But yes, probably better to check first if it's an actual string. And maybe
think if the object variants should even go through `fromTemplate` (it
might be that they're never template strings anyway).

ahocevar:
Thanks for the above link. Would you be able to update your pull request like I suggested above, or do you want me to make that change directly?

eliask:
Ah but `getValue()` cannot just always return strings.
There are various uses that require numbers, for instance (opacity, widths, etc.).

The correct return type of `getValue()` seems to be completely dependent on what is being retrieved.
To a first order approximation all the return types are given by:
https://github.com/mapbox/mapbox-gl-js/blob/master/src/style-spec/types.js#L197

E.g. `line-width` is `number | undefined`, from `"line-width"?: DataDrivenPropertyValueSpecification<number>`

Notable exceptions are the types in question in this pull request, which can/will(?) resolve to `ResolvedImage`, for instance:
```
export type ColorSpecification = string;
export type FormattedSpecification = string;
export type ImageSpecification = string;
```

Analogous processing for `mapbox-gl-js` always results in a `ResolvedImage` (for a symbol layer `icon-image`):
https://github.com/mapbox/mapbox-gl-js/blob/bfa913438ce0d65ddb38796d843dc1034569310d/src/data/bucket/symbol_bucket.js#L417

eliask:
I made a change to only apply `fromTemplate` when necessary.

It still mystifies me, however, that `fromTemplate` is necessary with the icons at all.
It seems that the template processing happens when there's an Mapbox GL style expresion to be evaluated. E.g. the following will return `ResolvedImage{ name: "test-11", ...}` for name=`test`.
```
'icon-image': {
  'base': 1,
  'stops': [
    [
      0,
      'circle-11'
    ],
    [
      8,
      '{name}-11'
    ]
  ]
```

*But* for simple expressions like `'icon-image': '{name}-11'`, it will just be a string `"{name}-11"`. It would be good to have this handling be done in a uniform manner, without ad hoc calls to `fromTemplate`.
Meanwhile, just merging this patch will fix the immediate problem.

ahocevar:
It may well be that the `fromTemplate()` call is not necessary at all. That was written before we started using Mapbox's expressions parser.

But for now your fix looks good, and I'm going to investiage the rest in a separate effort.

Thanks for your contribution!

PR #245, opened 2019-11-06
Parametrable credentials type

sylvainbeo:
@ahocevar I would do something like that. What do you think ?


ahocevar:
Thanks for this contribution! I think I'll prepare another pull request that introduces a general configuration object that can be passed to apply, applyStyle and the default export. One of the options will be the fetch credentials.

sylvainbeo:
@ahocevar Great !

ahocevar:
Fixed with #504.

PR #270, opened 2020-05-03
Type checking with imports from ol

ahocevar:
This pull request updates Typescript to v3.8. With this change, it will be possible to properly type check inheritance form OpenLayers, like it is done in #268.


package.json
```diff
@@ -30,13 +30,17 @@
   "dependencies": {
     "@mapbox/mapbox-gl-style-spec": "13.13.0",
     "mapbox-to-css-font": "^2.4.0",
+    "ol": "^6.3.2-dev.1588437998514",
```
tschaub, line 33: Was this intentionally added to `dependencies` instead of `devDependencies`?

ahocevar:
Oh, thanks for spotting this! It was meant to be under `devDependencies`.

ahocevar:
`ol` moved back to `devDependencies` with ac13bf5. Thanks again @tschaub!

PR #276, opened 2020-05-28
Make compatible with polyfill requirements for OpenLayers

mike-000:
Fixes https://github.com/openlayers/openlayers/issues/11103

Currently polyfills for Object.assign and String.prototype.startsWith are needed to use ol-mapbox-style on older browsers in addition to those required by OpenLayers.  olms already imports  `ol/obj#assign` via VectorLayer and VectorTileLayer and https://www.measurethat.net/Benchmarks/ListResults/3417 shows there is little if any performance advantage in using startsWith over indexOf for a single character on most browsers, and on Chrome it is considerably slower.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,8 +59,6 @@ fetch('data/states.json').then(function(response) {
 Internet Explorer (version 11) and other old browsers (Android 4.x) are supported when polyfills for the following features are loaded:
 
 -   `fetch` (including `Promise`)
--   `String.prototype.startsWith`
--   `Object.assign`
 
 ### Webpack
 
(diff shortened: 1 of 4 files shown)


ahocevar:
@mike-000 As you already noticed, this broke the legacy build. Any chance to fix it? Thanks in advance.

mike-000:
Do you have any idea of the cause?  It looks like the imported 'ol/obj' is undefined but I added it to the externals as you advised.

ahocevar:
`assign` from `ol/obj` is not part of the API, so it is not available in a legacy build.

mike-000:
I think an inline `object.assign` polyfill is used for workers somewhere in the `ol` source, but I cannot find it.  It could also fix this problem.

ahocevar:
If that is not part of the API, it won't be available in the legacy build either. The most pragmatic solution would be for ol-mapbox style to provide its own minimal polyfill.

PR #283, opened 2020-07-23
Fix undefined appended for single line < 35% width

M393:
Very short text with space like the house number '2 A' will get 'undefined' appended because both `prevWidth` and `nextWidth` are `Infinity` and the lines array is empty at that point.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -147,17 +147,20 @@ export function wrapText(text, font, em, letterSpacing) {
         lines.push(line);
       }
       // Pass 2 - add lines with a width of less than 30% of maxWidth to the previous or next line
-      for (let i = 0; i < lines.length; ++i) {
-        const line = lines[i];
-        if (measureText(line, letterSpacing) < maxWidth * 0.35) {
-          const prevWidth = i > 0 ? measureText(lines[i - 1], letterSpacing) : Infinity;
-          const nextWidth = i < lines.length - 1 ? measureText(lines[i + 1], letterSpacing) : Infinity;
-          lines.splice(i, 1);
-          if (prevWidth < nextWidth) {
-            lines[i - 1] += ' ' + line;
-            i -= 1;
-          } else {
-            lines[i] = line + ' ' + lines[i];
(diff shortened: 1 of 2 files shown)


src/util.js
```diff
@@ -147,17 +147,20 @@ export function wrapText(text, font, em, letterSpacing) {
...
-        if (measureText(line, letterSpacing) < maxWidth * 0.35) {
-          const prevWidth = i > 0 ? measureText(lines[i - 1], letterSpacing) : Infinity;
-          const nextWidth = i < lines.length - 1 ? measureText(lines[i + 1], letterSpacing) : Infinity;
-          lines.splice(i, 1);
-          if (prevWidth < nextWidth) {
-            lines[i - 1] += ' ' + line;
-            i -= 1;
-          } else {
-            lines[i] = line + ' ' + lines[i];
+      if (lines.length > 1) {
```
mikeydunnx, line 150: Will this error still occur if there are 2 lines and the last line triggers a wrap of an undefined?

ahocevar: @mikeydunnx What exactly would your test case look like?

mikeydunnx: I was thinking of that as well. It might be that this won't happen because short strings won't be wrapped.

However i still think its best to identify when an array item is undefined instead of assuming that this only happens when there is 1 line.

I had implemented a previous fix that appears to be missing for some reason.

https://github.com/openlayers/ol-mapbox-style/pull/256/files

ahocevar: @mikeydunnx That pull request raised another issue, so I had to modify that code again in #258. Maybe that whole function should be simplified or rewritten. In no step of the process it should create lines that are `undefined`, and it should update the array length so the last entry is not `undefined`.

M393:
@ahocevar, I think you forgot to push your test changes here from
https://github.com/openlayers/ol-mapbox-style/tree/text-do-not-append-undefined

ahocevar:
Thanks @M393, tests added with #285.

PR #304, opened 2020-10-01
Ensure there is a / at join in tile urls

mike-000:
This fixes incorrect tile urls generated by olms for the Esri OSM style https://www.arcgis.com/sharing/rest/content/items/3e1a00aeae81496587988075fe529f71/resources/styles/root.json
where neither the tile json path ends with / nor the tiles relative url starts with /

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -310,7 +310,7 @@ function setupVectorLayer(glSource, accessToken, url) {
         for (let i = 0, ii = tiles.length; i < ii; ++i) {
           const tile = tiles[i];
           if (tile.indexOf('http') != 0) {
-            tiles[i] = glSource.url + tile;
+            tiles[i] = glSource.url.replace(/\/?$/, '/') + tile.replace(/^\//, '');
           }
         }
       }


src/index.js
```diff
@@ -310,7 +310,7 @@ function setupVectorLayer(glSource, accessToken, url) {
         for (let i = 0, ii = tiles.length; i < ii; ++i) {
           const tile = tiles[i];
           if (tile.indexOf('http') != 0) {
-            tiles[i] = glSource.url + tile;
+            tiles[i] = glSource.url.replace(/\/?$/, '/') + tile.replace(/^\//, '');
```
ahocevar, line 313: Couldn't this be simplified?
```suggestion
            tiles[i] = glSource.url + tile.replace(/^\/?/, '/');
```

mike-000: Yes it would work but it would not have been needed if the TileJson url had been https://basemaps.arcgis.com/arcgis/rest/services/OpenStreetMap_v2/VectorTileServer/

Using `//` seems to be considered bad practice, even though it works in most cases? 

Avoiding `//` needs
`glSource.url.replace(/\/?$/, '/') + tile.replace(/^\//, '')`
or
`glSource.url.replace(/\/$/, '') + tile.replace(/^\/?/, '/')`

ahocevar: Ok, point taken. What about an empty `glSource.url`? Then we'd replace a relative url with an absolute url. Maybe also not desired?

mike-000: The code is inside `if (glSource.url) {` so relative tile urls always relative to the TileJson url unless a TileJson object is specified, in which case relative paths are relative to the application,

ahocevar: Alright, thanks for thinking that through with me.

ahocevar (approved):

PR #316, opened 2021-01-12
Correct import path in example snippet

mike-000:
Correct the `stylefunction` import path in the example snippet


ahocevar (approved):
Thanks, @mike-000

PR #341, opened 2021-08-17
fix: mapbox-gl-style-spec imports

ChristianKohler:
## Why?

Importing the `es.js` files directly breaks non es builds.

## Fix

Import by library name. That way it takes the appropriate module type.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -18,7 +18,7 @@ import View from 'ol/View.js';
 import applyStyleFunction, {getValue} from './stylefunction.js';
 import googleFonts from 'webfont-matcher/lib/fonts/google.js';
 import mb2css from 'mapbox-to-css-font';
-import {Color} from '@mapbox/mapbox-gl-style-spec/dist/index.es.js';
+import {Color} from '@mapbox/mapbox-gl-style-spec';
 import {assign, defaultResolutions} from './util.js';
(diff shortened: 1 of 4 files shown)


ahocevar:
Thanks, @ChristianKohler! If you'd be able to fix the linter error, that'd be great:
```
/home/circleci/repo/test/stylefunction-utils.test.js
  5:9  error  Replace `⏎··Color,⏎··latest·as·spec,⏎` with `Color,·latest·as·spec`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

ChristianKohler:
> Thanks, @ChristianKohler! If you'd be able to fix the linter error, that'd be great:
> 
> ```
> /home/circleci/repo/test/stylefunction-utils.test.js
>   5:9  error  Replace `⏎··Color,⏎··latest·as·spec,⏎` with `Color,·latest·as·spec`  prettier/prettier
> 
> ✖ 1 problem (1 error, 0 warnings)
>   1 error and 0 warnings potentially fixable with the `--fix` option.
> ```

Thank you @ahocevar for the quick response. Fixed the linting and test. Locally linting, test and build works.

I had to add the a path mapping in the test ts config to make it work.

ahocevar (approved):
Thanks, @ChristianKohler

PR #375, opened 2021-11-19
Include uid in functionCache index

mike-000:
Fixes https://github.com/openlayers/openlayers/issues/12862

Use the OpenLayers layer `uid` (or, for the background, the map `uid`) as an additional cache key for the `functionCache` and pass this in all calls to `getValue()`.  This will allow an application to have multiple maps (or a layer switcher) where the two or more different Mapbox styles share a tile set and therefore the same set of Gl layer ids.

There is a backward compatibility fallback to replicate the existing behaviour if `getValue()` is called externally without a `uid`.  However, subsequent to https://github.com/openlayers/openlayers/pull/13025 for background to work correctly with multiple layers once this is merged ol/layer/MapboxVector `configureSource()` should also be updated to pass `getUid(this)` to `getValue()` https://codesandbox.io/s/mapbox-vector-layer-forked-oi3w7?file=/main.js

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -11,6 +11,7 @@ import RenderFeature from 'ol/render/Feature.js';
 import Stroke from 'ol/style/Stroke.js';
 import Style from 'ol/style/Style.js';
 import Text from 'ol/style/Text.js';
+import {getUid} from 'ol/util.js';
 
 import mb2css from 'mapbox-to-css-font';
 import {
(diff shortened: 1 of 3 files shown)


ahocevar (changes requested):
Thanks for your finding the reason for the reported bug, and your attempt to fix this. I think a much simpler and more elegant fix would be to use the uid of the Mapbox style's `layer` as cache key. Then you wouldn't have to pass around arguments, and we would get cache invalidation for free when a user changes something in the Mapbox style and re-applies it.

mike-000:
I had not thought of using the `uid` of a non-`ol` object.  It is much simpler but as the broken test shows it will not work on a frozen style object.  If that is a problem is try/catch on `getUid()` and falling back to `layer.id` acceptable?

ahocevar:
Maybe caching the functions was an unnecessary over-optimization. But even if the optimization is necessary, it could maybe be better achieved by memoizing than by a cache with an insufficient cache key.

mike-000:
I am not sure what you mean by memorizing. If a single key is insufficient we need another. If the Gl style cannot be updated it rules out using the `uid` of the Gl layer, and the `ol` layer is the most obvious alternative.  To avoid repeatedly adding `uid` (along with` layer`, `zoom` and `f` which do not change) to 45 `getValue` calls a `getLayerValue` function could be defined within the `for (let i = 0, ii = layers.length; i < ii; ++i) {` loop which takes only `layoutOrPaint` and `property` and calls `getValue` with the remaining parameters.

ahocevar:
I was thinking about using `ol/functions/memoizeOne` for a `getFunction()` function with `layer` and `property` arguments that would replace `functionCache[layerId][property]`.

mike-000:
Cache is useful as it can be reset if `applyStyle` is called again with an updated Gl style.  Passing all the common parameters in one place makes the code tidier and easier to maintain.  `ol/layer/MapboxVector` will correctly set the background on multiple layers without any changes as an undefined `uid` uses a temporary cache, but passing the `ol` layer `uid` may be slightly more efficient as it will use the real cache.

ahocevar:
Thanks for your continued effort on this. Can you please add a test case that fails without your fix, and passes with the fix?

PR #427, opened 2022-01-10
Avoid applying background if layout visibility is none

tschaub:
In cases where a background layer has `"visibility": "none"` for the layout, the background color and opacity should not be applied.

It looks like the current code handles the case where a map (with a target element) is passed to `applyStyle` but not the case where a layer is passed.


ahocevar (approved):
Thanks for this fix,  @tschaub!

PR #434, opened 2022-01-19
Allow outline stroke for polygons without fill

ahocevar:
This pull request fixes the behavior when a fill-outline-color is set but no fill is rendered. In this case, the outline was previously not rendered. Now it is.

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -517,41 +517,48 @@ export default function (
               ),
               opacity
             );
-            if (color) {
-              if (layer.type + '-outline-color' in paint) {
-                strokeColor = colorWithOpacity(
-                  getValue(
-                    layer,
-                    'paint',
-                    layer.type + '-outline-color',
-                    zoom,
-                    f,
-                    functionCache
-                  ),
-                  opacity
-                );
-              }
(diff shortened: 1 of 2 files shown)


bartvde:
confirmed just now this fixes the issue

bartvde:
oh wait @ahocevar I am seeing something else now though, it seems it somehow does have a fill, in the default style or something?

See before and after:

Screenshot from 2022-01-19 17-16-19
Screenshot from 2022-01-19 17-16-12

ahocevar:
@bartvde I updated the pull request. Can you test again please?

bartvde:
one thing, but not sure if it's related though, that turning the layer off after turning the layer on doesn't get rid of it though, works fine for other layers though

bartvde:
turning off does work fine with the latest release and renderTransparent(true)

ahocevar:
Ok strange. This fix is supposed to work without setting `renderTransparent(true)`.

ahocevar:
Ok, I just added a test to make sure we're not doing anything wrong.

bartvde (approved):
LGTM, thanks for adding the test as well

PR #561, opened 2022-05-16
set declutter mode

cns-solutions-admin:
... based on icon-ignore-placement and icon-allow-overlap.

See https://github.com/openlayers/openlayers/issues/7475
See https://github.com/openlayers/openlayers/pull/13566

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -135,6 +135,49 @@ export function getValue(
 
 /**
  * @private
+ * @param {Object} layer Gl object layer.
+ * @param {string} allowOverlapProperty the allow-overlap property
+ * @param {string} ignorePlacementProperty the ignore-placement property
+ * @param {number} zoom Zoom.
+ * @param {Object} feature Gl feature.
+ * @param {Object} [functionCache] Function cache.
+ * @return {"declutter"|"obstacle"|"none"} Value.
+ */
+function getDeclutterMode(
+  layer,
+  allowOverlapProperty,
+  ignorePlacementProperty,
+  zoom,
+  feature,
+  functionCache
+) {
+  const allowOverlap = getValue(
+    layer,
+    'layout',
+    allowOverlapProperty,
+    zoom,
+    feature,
+    functionCache
+  );
+  if (!allowOverlap) {
+    return 'declutter';
+  }
+  const ignorePlacement = getValue(
+    layer,
+    'layout',
+    ignorePlacementProperty,
+    zoom,
+    feature,
+    functionCache
+  );
+  if (!ignorePlacement) {
+    return 'obstacle';
+  }
+  return 'none';
+}
+
+/**
+ * @private
  * @param {string} layerId Layer id.
  * @param {?} filter Filter.
  * @param {Object} feature Feature.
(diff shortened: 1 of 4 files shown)


src/stylefunction.js
```diff
@@ -975,6 +1027,16 @@ export function stylefunction(
...
+            // the mapbox style spec does not support decluttering circles.
+            // but as openlayers does, we can support it as "extension"
+            const declutterMode = getDeclutterMode(
+              layer,
+              'circle-allow-overlap',
+              'circle-ignore-placement',
+              zoom,
+              f,
+              functionCache
+            );
```
ahocevar, line 1039: Please do not add style properties that the spec does not support. Instead, there is something else to fix: Currently, we treat circles different than mapbox-gl-js, because we declutter them, which mapbox-gl-js does not. So a better change would be to change circle rendering to never declutter (`declutterMode: 'none'`).

When you change this, the signature of `getDeclutterMode()` will also be simpler, because it does not need the 2nd and 3rd argument any more.

ahocevar (changes requested):
Thanks for your work on this, @cns-solutions-admin. Please address my comment about circle decluttering. Then I'll give it another review.

src/stylefunction.js
cns-solutions-admin: The nice thing about the mapbox styles is the ability to configure the styles from a configuration file (which is the reason we want to use it for configuring openlayers layers, too).
However, there are openlayers features that are not supported by mapbox, like decluttering circles, for which it would be nice to have a configuration possibility, here with additional properties that don't hurt.

I would like to keep the properties, but can remove them, if you insist.
To confirm to the mapbox behaviour, the default for the circles will be declutter: none.

However, as before the PR, the style function only works correctly when applied to a layer that has declutter set correctly depending on the style's text-allow-overlap. When using apply for the whole map/mapbox style sheet, all layers are created with declutter = true. This is not solved with this PR.

ahocevar: The problem is: if you leave these undocumented properties in, they are not documented anywhere. I use ol-mapbox-style too for configuring styles from static objects, and for customisations I use layer metadata and a style function wrapper that modifies the returned styles. If `declutterMode` had a setter on `ol/style/Image`, this would be very simple in your case too. We could make it even simpler by providing a style customization hook, something like
```js
/**
 * @param {import("ol/style/Style.js").default} style Style.
 * @param {Object} glLayer Item from the Mapbox style's `layers` array.
 * @param {Object} glSource Item from the Mapbox style's `sources` object.
customizeStyle(style, glLayer, glSource) {
  if (glLayer.metadata.declutter-circle && style.getImage() instanceof CircleStyle) {
    style.getImage().setDeclutterMode(glLayer.metadata.declutter-circle)
  }
}
```
That function could become a new config option, to be configured in the `options` object.

src/stylefunction.js
```diff
@@ -989,6 +1032,7 @@ export function stylefunction(
                     color: circleColor,
                   })
                 : undefined,
+              declutterMode: declutterMode,
```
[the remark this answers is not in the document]
cns-solutions-admin: done

ahocevar (approved):
Thanks, @cns-solutions-admin

PR #564, opened 2022-05-18
Allow dynamic (non-sprite) images

cns-solutions-admin:
See https://github.com/openlayers/ol-mapbox-style/issues/563.

Please review, if this is an acceptable way to add the feature.


ahocevar:
I'm not in favor of this approach. See https://github.com/openlayers/ol-mapbox-style/issues/563#issuecomment-1129733750

ahocevar:
Ok, so let's go with this approach. I think we do not need the 2nd argument to the `getImage()` function, and we can keep track of image loading. For this, I have created https://github.com/CNS-Solutions/ol-mapbox-style/pull/1.

In addition, please add a unit test. And you'll probably want a way to pass a `getImage` function in the options of the `apply*` functions.

ahocevar:
@cns-solutions-admin Is this something you still want to work on?

cns-solutions-admin:
It seems I somehow missed your last comment and thought this PR was dead.
We definitely need this functionality and are currently using a patched ol-mapbox-style.

I will now continue work this week.

cns-solutions-admin:
@ahocevar: Included your suggestion to automatically trigger layer change, if image not yet loaded.
Also added getImage to apply/applyStyle options.
Tests added.

ahocevar:
Thanks for your continued effort on this, @cns-solutions-admin. I'll be giving it a proper review later this week.

ahocevar (commented):
Nice, thanks for your effort on this, @cns-solutions-admin. If the image function could return a `string` in addition to `HTMLImageElement|HTMLCanvasElement|undefined`, you could also return the `src` and thus avoid the overhead of creating a new Image element with the src of the returned image element that is not `complete` yet. And the documentation could add that when returning a `HTMLImageElement`, it has to be loaded already. Would you agree, and if so, would you be able to make this one final change?

Other than that, this looks great.

cns-solutions-admin:
Added string parameter and improved documentation.
We don't (and didn't) the load event, this should be handled by the Icon/IcomImage style.

ahocevar (approved):
Thanks, @cns-solutions-admin

PR #585, opened 2022-06-11
Simplify template handling and make it more secure

### LINKED ISSUE #584: Regex `const templateRegEx` in src/stylefunction.js vulerable to ReDoS
The regular expression `const templateRegEx = /^([^]*)\{(.*)\}([^]*)$/;` in `src/stylefunction.js` is vulnerable to the ReDoS attack according to several SCA tools including Sonarqube and [Devina.io](https://devina.io/redos-checker).

The RE is used by function `fromTemplate` which is called 3 times in `stylefunction.js` for various layer components including `fillIcon`, `iconImage`, and `textField`. While I do not have any PoC code yet, this is probably worth checking since layer info can theoretically be taken from external/unstrustworthy sources.

ahocevar:
Should fix #584.

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -229,7 +229,7 @@ function colorWithOpacity(color, opacity) {
   return color;
 }
 
-const templateRegEx = /^([^]*)\{(.*)\}([^]*)$/;
+const templateRegEx = /\{[^{}}]*\}/g;
 
 /**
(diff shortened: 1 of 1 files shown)


danielklim:
confirm this fixes #584 (though I haven't confirmed actual functionality... I assume your test suite confirms that).

PR #618, opened 2022-07-18
Call transformRequest for sprite image URLs

### LINKED ISSUE #617: applyStyle doesn't apply transformRequest to spriteImageUrl
Currently when the `transformRequest` option is passed to `applyStyle` it is used to transform URLs for requests of type Style, Sprite, Source and Tiles but the `spriteImageUrl` isn't passed to `transformRequest`.

I'd like to submit a PR which calls `transformRequest` (when defined) for the `spriteImageUrl` with a `type` of `SpriteImage` and includes a test to ensure `transformRequest` is called for each of the appropriate types for vector tile styles. I think I've got PR nearly ready but thought I'd better check before submitting it.

walkermatt:
Resolves #617 by calling `transformRequest` for the `spriteImageUrl` when `transformRequest` is passed as an option to `applyStyle`.

Tests verify that `transformRequest` is being called for each of the expected `ResourceType` values.

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -59,7 +59,7 @@ import {
  * @property {string} [accessTokenParam='access_token'] Access token param. For internal use.
  */
 
-/** @typedef {'Style'|'Source'|'Sprite'|'Tiles'|'GeoJSON'} ResourceType */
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @walkermatt!

PR #630, opened 2022-07-27
Use Object.assign

tschaub:
This makes use of `Object.assign` instead of importing `assign` from the `ol/obj.js` module.

See openlayers/openlayers#13888.

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -21,7 +21,6 @@ import {
   stylefunction as applyStyleFunction,
   getValue,
 } from './stylefunction.js';
-import {assign} from 'ol/obj.js';
 import {createXYZ} from 'ol/tilegrid.js';
 import {
   defaultResolutions,
@@ -70,7 +69,7 @@ import {
  */
 function completeOptions(styleUrl, options) {
   if (!options.accessToken) {
-    options = assign({}, options);
+    options = Object.assign({}, options);
     const searchParams = new URL(styleUrl).searchParams;
     // The last search parameter is the access token
     searchParams.forEach((value, key) => {
@@ -613,7 +612,7 @@ function setupGeoJSONSource(glSource, styleUrl, options) {
     });
   }
   const source = new VectorSource(
-    assign(
+    Object.assign(
       {
         attributions: glSource.attribution,
         format: geoJsonFormat,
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks!

ahocevar:
Feel free to merge even without the check run, I either need to switch to GitHub Actions or figure out why CircleCI does not pick up your pull request.

ahocevar:
I took the liberty to rebase your branch, because I have now configured GitHub Actions. Looks like the previously expected circleci check cannot be removed from existing pull requests, but with the Node.js CI check green this should be safe to merge.

tschaub:
I removed the CircleCI job from the required status checks under the [branch protection rules](https://github.com/openlayers/ol-mapbox-style/settings/branch_protection_rules/4398713).  And I added the new "build" job to the required status checks.

image

A couple only somewhat related things that you may or may not want to consider.  The names for the required status checks come from a concatenation of the job name in the yaml file and version from the strategy matrix.

So in this case, the status check will be prefixed with `build`:
https://github.com/openlayers/ol-mapbox-style/blob/d921f19b5db84b4c963da255aace74dae74e4948/.github/workflows/node.js.yml#L13

If you wanted it to be named `test` for example, that line would change.  I only mention this because I always get it mixed up with the `name` at the top of the workflow file - never knowing which is going to be the name in the required status check.

If you are only really concerned about testing against one version of Node, I think it is nicer to remove the strategy matrix and just use a fixed Node version.

```diff
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -14,17 +14,12 @@ jobs:
 
     runs-on: ubuntu-latest
 
-    strategy:
-      matrix:
-        node-version: [16.x]
-        # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
-
     steps:
     - uses: actions/checkout@v3
     - name: Use Node.js ${{ matrix.node-version }}
       uses: actions/setup-node@v3
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: '16.x'
         cache: 'npm'
     - run: npm ci
     - run: npm run build --if-present
```

That way, the required status check name is fixed (`build` in this case) and you don't have to edit your branch protection rules any time you update the Node version (which is a hassle because you can't add the new status check as required until it has been run).

ahocevar:
Thanks, @tschaub.

PR #634, opened 2022-07-28
Rename build job to test, remove version matrix

tschaub:
This renames the `build` job to `test` and removes the version matrix.  This means that the version can be updated without needing to edit the branch protection rules (the required status check name concatenates the job name and the version from the matrix).


tschaub:
If you agree with this change, the [branch protection rule](https://github.com/openlayers/ol-mapbox-style/settings/branch_protection_rules/4398713) can be edited to require the `test` status check instead of the `build (16.x)` status check.

ahocevar (approved):

tschaub:
I updated the branch protection rule to require the `test` status check.

image

PR #648, opened 2022-08-13
Apply to Layer Groups

scadergit:
Overall applying to layer group went as planned. I have three questions, marked in code with `???` comments. 
1. `// ??? Rename mapOrGroupOrLayer` Not sure readability or consistency is better here
2. `// ??? could apply to first layer` I might just need the weekend to think this over, but given background is often the first style layer processed, I can't simply apply to the first actual layer in the group
3. ` // ??? A group wouldn't have a reference` I assume this code still important even when used within a group, but not sure how to reference back to the map view when you start at a layer group

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -4,9 +4,7 @@ Copyright 2016-present ol-mapbox-style contributors
 License: https://raw.githubusercontent.com/openlayers/ol-mapbox-style/master/LICENSE
 */
 
-import Color from '@mapbox/mapbox-gl-style-spec/util/color.js';
 import GeoJSON from 'ol/format/GeoJSON.js';
-import LayerGroup from 'ol/layer/Group.js';
 import MVT from 'ol/format/MVT.js';
 import Map from 'ol/Map.js';
 import TileGrid from 'ol/tilegrid/TileGrid.js';
@@ -28,6 +26,7 @@ import {
   fetchResource,
   getGlStyle,
   getTileJson,
+  getZoomForResolution,
 } from './util.js';
 import {equivalent, fromLonLat, getUserProjection} from 'ol/proj.js';
 import {getFonts} from './text.js';
@@ -36,6 +35,8 @@ import {
   normalizeSpriteUrl,
   normalizeStyleUrl,
 } from './mapbox.js';
+
+/** @typedef {import("ol/layer/Group.js").default} LayerGroup */
 
 /**
  * @typedef {Object} FeatureIdentifier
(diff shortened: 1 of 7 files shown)


scadergit:
I included an example page named "Apply to Group"

src/apply.js
```diff
@@ -357,6 +358,7 @@ export function applyStyle(
 
 const emptyObj = {};
 
+// ??? Rename mapOrGroupOrLayer, or just leave as-is given Group is part of layer class?
```
ahocevar, line 361: No need to rename - a layer group is also a layer.

src/apply.js
```diff
@@ -412,11 +414,16 @@ function setBackground(mapOrLayer, layer) {
...
   }
   if (typeof mapOrLayer.getTargetElement === 'function') {
+    // map object
     if (mapOrLayer.getTargetElement()) {
       updateStyle();
     }
     mapOrLayer.on(['change:resolution', 'change:target'], updateStyle);
+  } else if (mapOrLayer instanceof LayerGroup) {
+    // group object
+    // ??? could apply to first layer, but probably doesn't exist yet
```
ahocevar, line 424: You could set it as property on the group (e.g. `group.set('mapbox-background', updateStyle)`. When you add the first layer to the group, pick it up from the group and set it on the layer.

src/apply.js
```diff
@@ -701,17 +719,20 @@ function processStyle(glStyle, map, styleUrl, options) {
...
-              this,
-              glLayer,
-              layer,
-              view,
-              functionCache
-            )
-          );
-          updateRasterLayerProperties(glLayer, layer, view, functionCache);
+          // ??? A group wouldn't have a reference to map/view?
+          if (view) {
```
ahocevar, line 723: This could be solved by listening to the layer's prerender event instead of the view's `change:resolution` event. That event receives the resolution as `event.frameState.viewState.resolution`, which would allow calling `updateRasterLayerProperties()` only when the resolution has changed.

ahocevar:
I think for now it's fine to leave the background handling the way it is. Better to improve it in OpenLayers than working around with a vector layer here.

src/apply.js
scadergit: It looks like the API has `setBackground` only on vector layers (for obvious reasons?), so I'll check for the first instance of one in the group.

src/apply.js
scadergit: Think I have this ready, but noticed that it seems background can't be applied to GeoJSON layer due to the following `updateStyle` function not accounting for it. This implies background can only be applied to `VectorTile` or `Map`? Is that by design or should I add it to `updateStyle`?
```
    const zoom =
      typeof mapOrLayer.getSource === 'function'
        ? mapOrLayer.getSource().getTileGrid().getZForResolution(resolution)
        : mapOrLayer.getView().getZoom();
```

src/apply.js
scadergit: I ahve this ready, but see comment to next conversation...

scadergit: Just noticed that a vector source doesn't have `setBackground` so looks like by design.

scadergit:
Got one last TS error that I'm perplexed over...as it works just fine. The error says I can't use a string of "change:source" but it's literally in the TS defnition.
image

ahocevar:
I'm going to commit a change to your branch that simplifies the background handling and makes the TypeScript problem (which comes from using `this` as context for `updateBackground()`) go away.

scadergit:
Ah, that makes sense...apprecate the help and look forward to your suggestions.

ahocevar:
@scadergit I added two commits. The first one is what I mentioned in my previous comment. It simplifies background handling by using a `prerender` hook to render the background with `fillRect()` to the first layer's canvas. According to the spec, the background is [applied below all other map content](https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#background). This is now also implemented correctly and can be seen in your new example.

The second commit only cleans up the example and changes its name.

if you agree with my changes, please mark the pull request as "Ready for review".

scadergit:
I like the switch to a prerender function, and thanks for catching that I left my Esri source in the example.

ahocevar (approved):
Thanks for the hard work on this, @scadergit

scadergit:
@ahocevar will this be released as 8.2.x or 9.x? Is there a release schedule or as-necessary?

ahocevar:
@scadergit as-necessary. And it will be in the next release (v9.1.0).

ahocevar:
v9.1.0 published.

ahocevar:
@scadergit We don't do backport releases, but if you're not using the `icon-offset` layout property (which is the only breaking change that requires OpenLayers 7.x), you can also use v9 with OpenLayers 6.x.

scadergit:
Had to bump npm to 8.3 so I could use overrides, but yay...
```
<template>
  <vl-layer-group @created="_created" />
</template>
...
<script>
    _created (group) {
      apply(group.$layer, this._styleUrlWithToken)
    }
</script>
```
image

PR #657, opened 2022-08-18
Icon offset updates for v7 and x-direction fix

### LINKED ISSUE #654: `icon-offset` moves left instead of right
The specification https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#layout-symbol-icon-offset says `Positive values indicate right and down`  However as implemented it moves left.  See https://codesandbox.io/s/print-to-scale-forked-cbx1cz?file=/main.js and comment out the `styleLayer.layout['icon-offset'] = [100, 100];` to see the original position of the station icon.  As noted in https://github.com/openlayers/openlayers/pull/13975#issuecomment-1215593028 this code would be affected by that issue, but currently `).map((v) => -v * spriteImageData.pixelRatio)` would need to `).map((v)(v, i) => (1 - 2 * i) * v * spriteImageData.pixelRatio)`.  If making a breaking change it might be worth making a further change to define displacement positive down consistent with `ol/style/Text` offsetY.

ahocevar:
Fixes #654.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## 9.0.0
+
+### Breaking changes
+
+* ol-mapbox-style now requires OpenLayers 7
+* `icon-offset` has been fixed. If you previously worked around the x-value moving left instead of right with increasing offset, you have to remove this workaround now.
+
+### Other changes
+
 ## 8.2.1
 
 * Updated dependencies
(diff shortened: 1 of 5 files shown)


mike-000:
Isn't that now going to move up instead of down?

mike-000:
No problem did not see the `displacement[1] *= -1;`

PR #663, opened 2022-08-26
apply is not default return

irvingvjuarez:


ahocevar:
Actually both should work. What was your build environment when you encountered this problem? But I agree that using the named export is nicer.

irvingvjuarez:
I was in dev env.
Vscode auto completed the import as a named import and it worked. Then I saw documentation as a default import and decided to make the change

ahocevar (approved):

PR #664, opened 2022-08-27
applyStyle method updated in docs. It works just the same way as apply

irvingvjuarez:
While using the library, I figured this out

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ import {applyStyle} from 'ol-mapbox-style';
 import VectorTileLayer from 'ol/layer/VectorTile.js'
 
 const layer = new VectorTileLayer({declutter: true});
-applyStyle(layer, 'mapbox://styles/mapbox/bright-v9', {accessToken: 'YOUR_MAPBOX_TOKEN'});
+applyStyle(layer, 'https://api.mapbox.com/styles/v1/mapbox/bright-v9?access_token=YOUR_MAPBOX_TOKEN');
 ```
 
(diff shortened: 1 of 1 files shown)


ahocevar:
We intentionally also show `mapbox://` URIs in the docs.  See a few lines above:
 > To assign style and source to a layer only, use `applyStyle()`. `mapbox://` urls are also supported:

README.md
```diff
@@ -35,7 +35,7 @@ import {applyStyle} from 'ol-mapbox-style';
 import VectorTileLayer from 'ol/layer/VectorTile.js'
 
 const layer = new VectorTileLayer({declutter: true});
-applyStyle(layer, 'mapbox://styles/mapbox/bright-v9', {accessToken: 'YOUR_MAPBOX_TOKEN'});
```
irvingvjuarez, line 38: Oh, alright. I made the PR because the `applyStyle` method is not working anymore with `(layer, [url], config)` as parameters. It works with `(layer, src, ...otherParams)` just as the `apply` method with map instead of layer.

ahocevar: Ok, can you provide an example of what does not work anymore please?

ahocevar: Ah, I get it now. Will create a pull request to make the advertised API work again.

ahocevar:
I have created #670, which adds some convenience to the `applyStyle()` signature - all options can now be provided in the 3rd argument, like shown in the docs, in the section that you changed.

irvingvjuarez:
Alright, thank you 👌

PR #672, opened 2022-09-01
Support build on windows

cns-solutions-admin:
Build/install on windows fails because of "esm=true". Fix by using cross-env.


ahocevar (approved):
Thanks, @cns-solutions-admin

PR #693, opened 2022-09-14
fix: mapbox satellite and streets v8

bovandersteene:
An issue occurs when trying to add a mapbox style with satellite. 
A lot of request are done to localhost/a, localhost/b, .... 

Also when the style includes a v8 of streets the tiles are not loaded. I compared how mapbox reads the implementation for the tiles and it just perform a call to api.mapbox.com instead of tiles.mapbox.com

### DIFF
--- a/src/mapbox.js
+++ b/src/mapbox.js
@@ -77,11 +77,9 @@ export function normalizeSourceUrl(url, token, tokenParam, styleUrl) {
     return decodeURI(urlObject.href);
   }
 
-  if(mapboxPath === 'mapbox.satellite'){
+  if (mapboxPath === 'mapbox.satellite') {
     const sizeFactor = window.devicePixelRatio >= 1.5 ? '@2x' : '';
     return `https://api.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}${sizeFactor}.webp?access_token=${token}`;
-
   }
   return `https://{a-d}.tiles.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}.vector.pbf?access_token=${token}`;
 }
-
--- a/src/mapbox.js
+++ b/src/mapbox.js
@@ -82,7 +82,6 @@ export function normalizeSourceUrl(url, token, tokenParam, styleUrl) {
     return `https://api.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}${sizeFactor}.webp?access_token=${token}`;
 
   }
-
-  return `https://api.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}.vector.pbf?access_token=${token}`;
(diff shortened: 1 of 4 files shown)


ahocevar (changes requested):
Thanks for your contribution, @bovandersteene. I added two remarks. Would be great if you could address them.

src/mapbox.js
```diff
@@ -76,5 +76,12 @@ export function normalizeSourceUrl(url, token, tokenParam, styleUrl) {
...
     return decodeURI(urlObject.href);
   }
-  return `https://{a-d}.tiles.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}.vector.pbf?access_token=${token}`;
+
+  if(mapboxPath === 'mapbox.satellite'){
+    return `https://api.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}@2x.webp?access_token=${token}`;
+
+  }
+
+  return `https://api.mapbox.com/v4/${mapboxPath}/{z}/{x}/{y}.vector.pbf?access_token=${token}`;
```
ahocevar, line 85: According to [Mapbox documentation], the url should look like https://a.tiles.mapbox.com/v4/mapbox.mapbox-streets-v8/14/4823/6160.mvt?access_token=...

bovandersteene: I just investigated into the reason why I got this issue, and solved it. For a vector I reverted the changes for this url

src/util.js
```diff
@@ -117,7 +117,7 @@ export function getTileJson(glSource, styleUrl, options = {}) {
         promise = Promise.resolve(
           Object.assign({}, glSource, {
             url: undefined,
-            tiles: normalizedSourceUrl,
+            tiles: glSource.type === 'raster' ? [normalizedSourceUrl] : normalizedSourceUrl,
```
ahocevar, line 120: Shouldn't the ternary expression be something like
```js
Array.isArray(normalizedSourceUrl) ? normalizedSourceUrl : [normalizedSourceUrl]
```
?

ahocevar (changes requested):
I added one more comment, and please also address the lint issues found by the CI.

src/util.js
bovandersteene: I think I don't get it why we should check if the normalizedSourceUrl is an array.

While testing the mapbox style where I had the issue with, it turned out that the implementation inside setupRasterLayer expects an array for the tiles. For setupVectorLayer it doesn't except an array.

ahocevar: Ok, thanks for investigating that. I thought it always expects an array, which would be consistent. Then I'll make that consistent in a separate pull request.

ahocevar (approved):
Thanks for your effort, @bovandersteene

PR #754, opened 2022-12-06
Fix bbox parsing when using transformRequest

### LINKED ISSUE #753: bbox transform is broken with transformRequest
In wfs bbox transform is broken when using transformRequest:
`
apply('map', 'data/geojson-wfs.json',{transformRequest:((url, type) =>{return new Request(url)})});
`
I will provide a pull request...

AlenKelemen:
Fixes #753

### DIFF
--- a/examples/geojson-wfs.js
+++ b/examples/geojson-wfs.js
@@ -1,4 +1,12 @@
 import 'ol/ol.css';
 import {apply} from 'ol-mapbox-style';
 
-apply('map', 'data/geojson-wfs.json');
+apply('map', 'data/geojson-wfs.json', {
+  transformRequest: (urlStr, type) => {
+    if (type === 'GeoJSON') {
+      const url = new URL(urlStr + '&transformRequest=true');
+      const req = new Request(url);
+      return req;
+    }
+  },
+});
--- a/src/apply.js
+++ b/src/apply.js
@@ -653,12 +653,6 @@ function setupGeoJSONSource(glSource, styleUrl, options) {
       options.accessTokenParam || 'access_token',
       styleUrl || location.href
     );
-    if (options.transformRequest) {
-      const transformed = options.transformRequest(geoJsonUrl, 'GeoJSON');
-      if (transformed instanceof Request) {
-        geoJsonUrl = encodeURI(transformed.url);
-      }
-    }
     if (geoJsonUrl.indexOf('{bbox-epsg-3857}') != -1) {
       const extentUrl = (extent) => {
         return geoJsonUrl.replace(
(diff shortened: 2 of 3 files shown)


ahocevar:
Thanks, @AlenKelemen. Can you please elaborate on what exactly was broken, and what this fixes? Ideally by adding a test that fails before and passes after your change? To me it looks like your change just makes it so the original url will be used now, instead of the transformed one. So to properly fix this, `transformRequest` would have to be called in the url function, in addition to the current calls site.

AlenKelemen:
For mapbox geojson source one can use WFS request with bbox as:
```
https://ahocevar.com/geoserver/wfs?service=WFS&version=1.1.0&request=GetFeature&typename=osm:water_areas&outputFormat=application/json&srsname=EPSG:4326&bbox={bbox-epsg-3857}
```
where flag bbox={bbox-epsg-3857} must be replaced in code with real bounding box coordinates.

So we added appropriate code in apply.js:

``` 
if (geoJsonUrl.indexOf('{bbox-epsg-3857}') != -1) {
      const extentUrl = (extent) => {
        return geoJsonUrl.replace(
          '{bbox-epsg-3857}',
          `${extent.join(',')},EPSG:3857`
        );
      };
      const source = new VectorSource({
        attributions: glSource.attribution,
        format: geoJsonFormat,
        url: extentUrl,
        strategy: bboxStrategy,
      });
      source.set('mapbox-source', glSource);
      return source;
    } 
```

**What was broken:**

In function 'apply' one of the options is transformRequest which returns Request.
Request object encodes {bbox-epsg-3857} to %7Bbbox-epsg-3857%7D .
So in that form it slips: `if (geoJsonUrl.indexOf('{bbox-epsg-3857}') != -1){...}`
and bbox={bbox-epsg-3857} is never converted to coordinates, causing error.

**Solution:**
You are right about changing order in my pullRequest, it misses transformRequest.

So I suggests to replace `geoJsonUrl = encodeURI(transformed.url)` with `geoJsonUrl = decodeURI(transformed.url)`.
This way we get decoded url regardless of transformRequest option.
Or we can test if %7Bbbox-epsg-3857%7D (encodedUrl) as you suggests...

I added test for transformRequest..

examples/geojson-wfs.js
```diff
@@ -1,4 +1,12 @@
...
-apply('map', 'data/geojson-wfs.json');
+apply('map', 'data/geojson-wfs.json', {
+  transformRequest: (urlStr, type) => {
+    if (type === 'GeoJSON') {
+      const url = new URL(urlStr + '&transformRequest=true');
+      const req = new Request(url);
+      return req;
+    }
+  },
+});
```
ahocevar, line 12:
```suggestion
apply('map', 'data/geojson-wfs.json');
```

ahocevar, line 12:
```suggestion
apply('map', 'data/geojson-wfs.json');
```

ahocevar (changes requested):

examples/geojson-wfs.js
ahocevar: This change does not make sense in the example, and you have covered it in the test already. So please revert this example change.

ahocevar:
I'm going to merge this now, and apply my suggestion in a separate commit.

PR #761, opened 2022-12-13
bugfix: pending requests metadata

ejn:
fetchResource() caches pending requests (or rather the Promises of their
responses) to prevent repeated calls for the same resource. If a
response is not pending then the generated Request object may be
returned via the metadata parameter, but before this commit then if the
response was already pending the Request object was never returned in
this way. This led to problems in getTileJson() where it is expected
(required) that the request has been set.

With this PR then not just the Promises of the responses but also the
original Request objects are cached so that the request parameter can
also be set in the metadata object when a pending response is returned.

The problem was noticed while trying to use apply() in parallel on two
group layers using each a different style URL, which however both
used the same tileset as source.

### DIFF
--- a/src/util.js
+++ b/src/util.js
@@ -53,7 +53,10 @@ const pendingRequests = {};
  */
 export function fetchResource(resourceType, url, options = {}, metadata) {
   if (url in pendingRequests) {
-    return pendingRequests[url];
+    if (metadata) {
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks for the fix and the tests!

PR #795, opened 2023-02-03
Make applyStyle respect existing VectorLayer source

jnystad:
This fix makes applyStyle respect a pre-existing source on VectorLayer. As `applyStyle` docs claim, it shall not override a source if it is already set, but the check for equality on `mapbox-source` property triggers a source update if the source set was not created from the mapbox style specification itself.

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -298,7 +298,13 @@ export function applyStyle(
           }
           const glSource = glStyle.sources[sourceId];
           let source = layer.getSource();
-          if (!source || source.get('mapbox-source') !== glSource) {
+          if (
+            !source ||
+            (
+              source.get('mapbox-source') &&
+              source.get('mapbox-source') !== glSource
+            )
+          ) {
             source = setupGeoJSONSource(glSource, styleUrl, options);
           }
           const targetSource = /** @type {VectorSource} */ (layer.getSource());


ahocevar (changes requested):
Thanks for your contribution, @jnystad. Would you be able to add a test please?

jnystad:
Yes, but I need advice, I see. This change actually breaks a test case (line 77 in applyStyle.test.js) where it is expected for applyStyle to update the source with the url from the source defined in the mapbox style file.

Since this test is here, it seems there might be expectations that a pre-existing source will be configured, which is undocumented behaviour, and a breaking change. Any thought regarding this? It was added in #520 

Note that this breaks one of our intended use cases for this library, where we have a fully pre-configured source on a vector layer and only want to apply the style from the mapbox style file without any modifications to the source at all.

jnystad:
Actually, it is documented behaviour, on the `layer` parameter itself.

I think this PR should be closed, or perhaps be changed to add an option for this behaviour. Any advice would be much appreciated.

ahocevar:
What if you just configure your existing source accordingly, e.g.
```js
layer.getSource().set('mapbox-source', 'my-source');
applyStyle(layer, 'my-source');
```
or
```js
const mySource = new VectorSource({
  // ... your config
  properties: {
    mapbox-source: 'my-source'
  }
});
layer.setSource(mySource);
applyStyle(layer, 'my-source');
```
? To me this looks like a sensible way to achieve what you want.

ahocevar:
Also, I'm not sure why your application breaks, because I think this example also does what you want to do, and it works fine:
https://openlayers.org/en/main/examples/vector-tiles-4326.html

jnystad:
To your first suggestion, it checks against the source object, not the identifier, I believe.

It breaks because we have an invalid source config. I certainly realize our case is not "common", but it can also be argued that it triggers a load of a source that is not necessarily used at all (and updates parameters like URL and attribution), so I still would like a way to achieve the "pure" `applyStyle` behaviour.

I'm committing a suggested new option, although I'm not sure it is an option you would prefer to add.

ahocevar:
@jnystad You're right, my bad. The check is against the source object, not the identifier.

jnystad:
Do you think this option could be added? Could perhaps rename it to `applySource` or something else.

ahocevar:
I could also think about it the other way around: an `source: Source` option, so instead of configuring the source on the layer, you configure it in the `source` option. What do you think?

jnystad:
That is probably also fine, but removes the convenience it has today and is also a breaking change. I would normally go for a less intrusive change if possible, and adding an option to ignore source seems like a minor addition, but it is of course up to you.

ahocevar (approved):

ahocevar:
Convinced. Thanks, @jnystad

jnystad:
Thanks!

PR #809, opened 2023-02-23
Update mapbox.js

L-hikari:
Hi~! in examples/mapbox when user do not input anything, 'key' will be null, then refresh the brower the alert will not display.


ahocevar (approved):
Thanks, @L-hikari

PR #845, opened 2023-03-14
add `removeMapboxLayer` to exports

simonseyock:
This adds the missing export `removeMapboxLayer` to the exports of index.js


ahocevar:
Published as v10.3.2.

PR #875, opened 2023-03-31
Fix icon offset flip

### LINKED ISSUE #874: re-applying a "symbol" style causes icon-offset to flip
The usage of `updateMapboxLayer` causes the Y-component of an `icon-offset` of a mapbox-style `symbol`-layer to flip.

Link to reproduction-repo:
https://github.com/RobertOrthofer/ol-flipping-icons

RobertOrthofer:
This fixes #874, preventing the icon-offset to change after re-applying the same style.

This is opened asa Draft-PR, as the test does not seem to fail in the current main, most likely because the rendering is not triggered correctly.

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -878,7 +878,7 @@ export function stylefunction(
                         f,
                         functionCache,
                         featureState
-                      );
+                      ).slice(0);
                       displacement[1] *= -1;
                     }
                     const color = iconColor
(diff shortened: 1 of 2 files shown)


ahocevar (approved):
Thanks, @RobertOrthofer. I took the liberty to fix the test, and reword your commits. Note that we don't use machine readable commit message prefixes. Instead, we start the commit message with an uppercase letter.

PR #882, opened 2023-04-08
Add a working Terrarium example

mike-000:
The previous example had multiple problems (as well as needing a SKU token). 

The Mapbox tiles used Mapbox encoding - decoding them as Terrarium would produce unrealistic elevation values - Terrarium decoding needs a real Terrarium  source.

Also the Mapbox tiles had a buffer/gutter which is auto-detected by Mapbox when the source type is `raster-dem`.  For WebGL a gutter is used to provide offset pixels at the edge of textures preventing incorrect shading making the edges visible.  It is not needed for `ol/source/Raster` but the source needs to be aware of it to render correctly, and it cannot be deduced from the style, which is a compatibility issue (therefore the example uses a Terrarium source without gutter even though sources with gutter are available).

Going forward I think OpenLayers could auto-detect a DEM gutter as a `TileJSON`/`XYZ` source option, in which case `gutter` would then need to be set as a property of the tile for use in rendering.


ahocevar (approved):
Thanks, @mike-000.

Automatic detection of gutters is a separate topic and could be solved in OpenLayers by comparing natural image width/height with the specified tile size.

PR #920, opened 2023-05-26
Create layer for added mapbox layer

### LINKED ISSUE #917: Cannot add mapbox layer if source was not referenced before
Suppose you apply a style to a map with a mapbox source that has no associated mapbox layer in the beginning. Later you add a layer with `addMapboxLayer`, but you will hit this lines:

https://github.com/openlayers/ol-mapbox-style/blob/a3c64e620358fc6494941e3d088e343d17089b78/src/util.js#L453-L455

This can never be true as there exists no layer to this source. The question arises why this check is there in the first place? Why do the layers have to be inserted in this particular order? It would make it easier for the user if the layers could be inserted in any order.

If the order really is important, then the check should look like this instead:
```
if (mapboxLayers[index - 1].source !== mapboxLayer.source && mapboxLayers.some(layer => layer.source === mapboxLayer.source)) {
```
Also it might be nice that if `lastID` was undefined that the function would automatically search for the last layer using this source.

And there needs to be created a new openlayers source. I am not exactly sure where this will have to happen.

ahocevar:
This pull request removes the restriction that `addMapboxLayer()` can only add a layer when it has the same source as an adjacent layer.

Fixes #917.

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -826,6 +826,93 @@ function manageVisibility(layer, mapOrGroup) {
   onChange();
 }
 
+export function setupLayer(glStyle, styleUrl, glLayer, options) {
+  const functionCache = getFunctionCache(glStyle);
+  const glLayers = glStyle.layers;
+  const type = glLayer.type;
+
+  const id = glLayer.source || getSourceIdByRef(glLayers, glLayer.ref);
+  const glSource = glStyle.sources[id];
+  let layer;
+  if (type == 'background') {
+    layer = setupBackgroundLayer(glLayer, options, functionCache);
+  } else if (glSource.type == 'vector') {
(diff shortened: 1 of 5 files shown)


src/apply.js
```diff
@@ -826,6 +826,93 @@ function manageVisibility(layer, mapOrGroup) {
...
+    });
+    layer.setVisible(
+      glLayer.layout ? glLayer.layout.visibility !== 'none' : true
+    );
+  }
+  const glSourceId = id;
+  if (layer) {
+    layer.set('mapbox-source', glSourceId);
+  }
+  return layer;
```
simonseyock, line 913:
```suggestion
  glLayers.push(glLayer);
  return layer;
```

PR #939, opened 2023-06-29
Fix @mapbox/mapbox-gl-style-spec version to 13.17.1

petrsloup:
There is a possible legal problem with `ol-mapbox-style` as it depends on `@mapbox/mapbox-gl-style-spec` `^13.23.1`. The source code of this package is actually part of the `mapbox-gl-js` repo (https://github.com/mapbox/mapbox-gl-js/tree/main/src/style-spec). The code in this repo is now restricted to be used only with mapbox services.
It is possible that this makes `ol-mapbox-style` legally usable only with mapbox services. (And possibly also the whole `ol` ?)

This PR changes the dependency to the latest version before the license change (`v13.17.1`).

Alternatively, `@maplibre/maplibre-gl-style-spec` `v14.0.2` could be used (https://github.com/maplibre/maplibre-style-spec/blob/main/CHANGELOG.md#1402) but that would require more changes in the code and I'm not sure it's actually 100 % compatible.

Also, there is an issue with circular dependency: `ol-mapbox-style@main` -> `ol@^7.3.0` -> `ol-mapbox-style@^9.2.0` -> `@mapbox/mapbox-gl-style-spec@^13.23.1`.
So this PR ensures that `ol-mapbox-style` does not use any code from the newer version, but it is still installed as a dependency, until the dependency in `ol` is upgraded and `ol` is released and dependency in `ol-mapbox-style` is upgraded 😕 

@ahocevar Any better ideas how to solve the issue?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "10.6.0",
       "license": "BSD-2-Clause",
       "dependencies": {
-        "@mapbox/mapbox-gl-style-spec": "^13.23.1",
+        "@mapbox/mapbox-gl-style-spec": "13.17.1",
         "mapbox-to-css-font": "^2.4.1",
         "ol": "^7.3.0"
       },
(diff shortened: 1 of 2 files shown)


ahocevar:
@petrsloup The @mapbox/mapbox-gl-style-spec package is clearly published under the ISC license. See https://www.npmjs.com/package/@mapbox/mapbox-gl-style-spec or https://cdn.jsdelivr.net/npm/@mapbox/mapbox-gl-style-spec/package.json.

Even if the above is by mistake, the repository license (https://github.com/mapbox/mapbox-gl-js/blob/main/LICENSE.txt) clearly states that it only applies to the Mapbox Web SDK, which the style spec is not a part of. In that license file, it is also clearly stated that the repository "Contains code from mapbox-gl-js v1.13 and earlier", along with a copy of the BSD 3-clause license which applies to that code.

So to me it does not look like we need to change anything here.

petrsloup:
Ok, thanks for the clarification. I believe it makes sense and that everything is ok.

Issue #10, opened 2017-01-10
Patched Mapbox GL Spec


lukasmartinelli:
@ahocevar Would it make sense to have a patched Mapbox GL spec that only contains the properties and layer types that are supported by `ol-mapbox-style`?

I would take care of it. It could be in this repo or in another one.

For https://github.com/maputnik/editor this would allow me to only display properties for a OL3 style that can be supported.
And in general it would help to specify and document the subset of the Mapbox GL spec that is supported by `ol-mapbox-style`.

As always - I love this plugin.

ahocevar:
Are you talking about e.g. https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/reference/v8.json? In general, I like the idea. I'm just wondering what would be the best way to implement it. Ideally, unsupported features would just be commented out, but JSON does not allow comments. On the other hand, if unsupported features are removed from the spec, it will be harder to add them back later. Do you have any thoughts on this?

lukasmartinelli:
> Are you talking about e.g. https://github.com/mapbox/mapbox-gl-style-spec/blob/mb-pages/reference/v8.json?

Yes.

> In general, I like the idea. I'm just wondering what would be the best way to implement it. Ideally, unsupported features would just be commented out, but JSON does not allow comments. On the other hand, if unsupported features are removed from the spec, it will be harder to add them back later. Do you have any thoughts on this?

Perhaps fork and add ol-mapbox-style the `sdk-support` field. This way we only annotate properties that are supported and do not delete/add anything to the spec.

```json
          "sdk-support": {
            "basic functionality": {
              "js": "0.10.0",
              "android": "2.0.1",
              "ios": "2.0.0",
              "macos": "0.1.0"
            }
```

ahocevar:
So you mean something like

          "sdk-support": {
            "basic functionality": {
              "js": "0.10.0",
              "android": "2.0.1",
              "ios": "2.0.0",
              "macos": "0.1.0",
              "ol-mapbox-style: "v1.0.0"
            }

I agree that doing this is the cleanest way. If you agree, before creating a fork, I'd create an upstream ticket with the suggestion - maybe it can even be added there.

lukasmartinelli:
> I agree that doing this is the cleanest way. If you agree, before creating a fork, I'd create an upstream ticket with the suggestion - maybe it can even be added there.

Good idea, try that. I think the chances are not so big but upstream would be the best way.

[...]

PR #940, opened 2023-06-30
Use maplibre instead of mapbox style spec

ahocevar:
I'm putting this up for discussion, as an alternative to #939.

The one thing that makes me hesitant to make this change is that I don't know what the plans are for Maplibre to stick with or depart from the Mapbox style spec. If someone from @maplibre, maybe @birkskyum, could comment on that, it would make the decision easier.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,7 +14,6 @@
     "import/resolver": {
       "alias": {
         "map": [
-          ["@mapbox/mapbox-gl-style-spec", "./node_modules/@mapbox/mapbox-gl-style-spec"],
           ["ol-mapbox-style", "./src"],
           ["ol-mapbox-style/dist/stylefunction", "./src"]
         ],
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -490,7 +490,7 @@ ol-mapbox-style now requires OpenLayers >= 6.13.
 
 ## 6.1.1
 
-* Fix version issue with @mapbox/mapbox-gl-style-spec
+* Fix version issue with @maplibre/maplibre-gl-style-spec
 
 ## 6.1.0
 
--- a/eslint.config.js_
+++ b/eslint.config.js_
@@ -1,0 +1,36 @@
+import openlayers from 'eslint-config-openlayers';
+
+/**
+ * @type {Array<import("eslint").Linter.Config>}
+ */
+export default [
+  ...openlayers,
+  {
+    name: 'test-config',
+    files: ['test/**/*'],
+    languageOptions: {
+      globals: {
+        after: 'readonly',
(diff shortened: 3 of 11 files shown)


birkskyum:
Sure, since the Mapbox style spec has become proprietary, the plan for MapLibre is to continue developing the [MapLibre Style Spec](https://maplibre.org/maplibre-style-spec/) in collaboration with our partners independently from Mapbox. The main divergence has so far been around atmosphere / globe features which were added after the fork, but the two style specs are still largely identical, and most changes happen as backwards-compatible additions.

mbeckem:
@ahocevar It appears that new versions of `@mapbox/mapbox-gl-style-spec` will also fall under the proprietary license as of this commit: https://github.com/mapbox/mapbox-gl-js/commit/a02c71b7ebacbbe6ccf4279edd6dc2399f85a995. So switchting to MapLibre Style Spec seems like a good idea moving forward.

ahocevar:
@mbeckem Thanks for the pointer! I'll do this first, and see what the options are:
> If you are interested in using this module in a manner not expressly permitted by its license, please do not hesitate to contact legal@mapbox.com with details of what you have in mind.

ahocevar:
Since I got no satisfying answer from Mapbox, I'm inclined to go in that direction too. But it's also tempting to create a new package, independent of both Mapbox and Maplibre, with a few more useful features (profiles, auto-generated docs, optimized for build size).

orangemug:
@ahocevar the spec is in progress over at https://github.com/maparatus/ol-mapbox-style-spec, rendered spec at https://maparatus.github.io/ol-mapbox-style-spec/ (I'll continue to update issue #10 on progress).

That's based on the maplibre-gl spec. I'm happy to transfer that repository over once I complete the tests and update the spec.

From my perspective. It'd be fine to merge this in the meantime, as it'd be closer to the incoming spec work.

ahocevar:
Finishing this now, mostly to get rid of the extra tooling needed for flow types and stripping out unneeded properties from the spec.

HarelM:
If there's anything we can help with, please ping me as I'm the lead maintainer for maplibre-gl-js and maplibre-style-spec for the js part of it.
Any help with PRs and maintenance would be very much welcome.
I'm also maintaining (on my free time) Maputnik, which has a way to switch to open-layer renderer, any help there would also be much appreciated. This tool is a good way to test things for maplibre-gl-js, I guess the same is true for this package as well.

ahocevar:
Thanks for your offer, @HarelM. Much appreciated. Not sure I'll be able to help much with what you're asking for though - I have a bunch of packages to maintain myself, and doing this all in my free time.

PR #947, opened 2023-07-14
Add support for SDF sprites

petrsloup:
This PR adds support for SDF sprites.

Though not very well documented, the mapbox-gl-js has support SDF sprites for years.
Such sprites contain "distances" instead of the regular color information:
image
In such cases, the `sprite.json` has a `sdf: true` on each image.

The benefits of the SDF sprites include better scalability and ability to recolor the images nicely.
We would like to start using SDF sprites in MapTiler in the future, but still remain compatible with the `ol+ol-mapbox-style` stack, so this PR adds a basic support for the SDF sprites.

- In case of regular icons, the whole sprite image atlas is "unSDFied" once when it's first needed and then used the regular way (color tinting works the same as before)
- In case of halo, only the specific icon is "unSDFied" with proper coloring and halo color. (Color tinting would also tint the halo, so the color needs to be burned in.)
- SDF sprites are not supported in patterns, the is the case even in the mapbox specification.

There is no performance change for non-SDF sprites at all.


ahocevar (approved):
Thanks, @petrsloup!

PR #1009, opened 2023-10-27
New 'updateMapboxSource()' utility function

### LINKED ISSUE #1008: Make source available in transformRequest
It would be convenient to have the source available in the transformRequest call. I would like to set cql filters on some sources dynamically.

In older versions I was able to change the url of the source directly, but this is not possible anymore, as only loaders are used.

ahocevar:
Fixes #1008.

**updateMapboxSource**(`mapOrGroup`, `id`, `mapboxSource`): `Promise`&lt;`Source`>

Updates a Mapbox source object in the style. The according OpenLayers source will be replaced
and the map will be re-rendered.

##### Parameters

| Name           | Type                  | Description                                        |
| :------------- | :-------------------- | :------------------------------------------------- |
| `mapOrGroup`   | `Map` \| `LayerGroup` | The Map or LayerGroup `apply` was called on.       |
| `id`           | `string`              | Key of the source in the `sources` object literal. |
| `mapboxSource` | `any`                 | Mapbox source object.                              |

##### Returns

`Promise`&lt;`Source`>

Promise that resolves when the source has been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ and open a browser on the host and port indicated in the console output (usually
 - [setFeatureState](#setFeatureState)
 - [stylefunction](#stylefunction)
 - [updateMapboxLayer](#updateMapboxLayer)
+- [updateMapboxSource](#updateMapboxSource)
 
 ### References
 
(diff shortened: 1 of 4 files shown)


ahocevar:
@simonseyock, do you want to see if this supports your use case?

simonseyock:
Sorry for not responding, I will test this in this week.

simonseyock:
@ahocevar I finally was able to test this. It works really nice, but there is one problem: Sometimes I would want to update a source and a layer at the same time. This causes two renders, where one is in an incomplete state.

I have ways to circumvent this, and probably do not have time in the project anymore to test any proper solutions for this problem. Wanted to report my findings anyways.

PR #1031, opened 2023-11-21
Apply icon-size (scale) for imageElement icons

jnystad:
This adds the missing scale property on Icon when using non-sprite images.


ahocevar (approved):
Thanks, @jnystad

PR #1043, opened 2023-11-30
tilegrid resolutions can use options

L-hikari:
Hi~ When I use `applyStyle` function with custom resolutions, I find the VectorTileSource tilegrid resolutions not use my custom. So I update this.


ahocevar:
Thanks for your attempt to fix something! Unfortunately the custom resolutions are not meant to configure the tile grid. They are only for mapping the zoom levels of the style to the appropriate resolutions.

If you're working with a source that does not use a standard tilegrid, you need to use `applyStyle()` on a layer that has a source already with the appropriate tile grid.

L-hikari:
Thanks for your reply! I had use a new source with my custom resolutions.

PR #1053, opened 2023-12-12
Allow default `line-color` when property not present

orangemug:
Previously the `default` from the property spec for `line-color` was not being applied because of the clause `'line-color' in paint`

This PR fixes that, which means the following layer style will show a line (in line with maplibre/mapbox)

```json
{
    "id": "test",
    "type": "line",
    "source": "lines",
    "layout": {},
    "paint": {}
}
```


ahocevar (approved):
Thanks, @orangemug

PR #1054, opened 2023-12-12
Allow `background-color` to use default from spec

orangemug:
This PR is a fix to use the default from the spec for the `background-color`. The following now results in a `#000000` background as per the spec.

```json
{
    "id": "background",
    "type": "background",
    "paint": {}
}
```


ahocevar (approved):
Thanks, @orangemug

PR #1055, opened 2023-12-12
Added `raster-`(`hue-rotate`/`contrast`/`saturation`/`opacity`/`brightness-*`/) as operations

orangemug:
This PR adds the following styling rules

 - [`raster-hue-rotate`](https://docs.mapbox.com/style-spec/reference/layers/#paint-raster-raster-hue-rotate)
 - [`raster-contrast`](https://docs.mapbox.com/style-spec/reference/layers/#paint-raster-raster-contrast)
 - [`raster-saturation`](https://docs.mapbox.com/style-spec/reference/layers/#paint-raster-raster-saturation)
 - [`raster-brightness-min`](https://docs.mapbox.com/style-spec/reference/layers/#paint-raster-raster-brightness-min)
 - [`raster-brightness-max`](https://docs.mapbox.com/style-spec/reference/layers/#paint-raster-raster-brightness-max)

Missing: Tests for new `raster-*` operations.

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -716,9 +716,21 @@ function setupRasterLayerAbstract(glSource, styleUrl, options) {
  * @param {Object} glSource "source" entry from a Mapbox Style object.
  * @param {string} styleUrl Style url
  * @param {Options} options ol-mapbox-style options.
- * @return {ImageLayer<Raster>} The raster layer
+ * @return {TileLayer} The raster layer
  */
 function setupRasterLayer(glSource, styleUrl, options) {
+  const tileLayer = setupRasterLayerAbstract(glSource, styleUrl, options);
+  return tileLayer;
+}
+
+/**
+ *
+ * @param {Object} glSource "source" entry from a Mapbox Style object.
+ * @param {string} styleUrl Style url
+ * @param {Options} options ol-mapbox-style options.
+ * @return {ImageLayer<Raster>} The raster layer
+ */
+function setupRasterOpLayer(glSource, styleUrl, options) {
   const tileLayer = setupRasterLayerAbstract(glSource, styleUrl, options);
   /** @type {ImageLayer<Raster>} */
   const layer = new ImageLayer({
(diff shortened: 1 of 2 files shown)


src/apply.js
```diff
@@ -1033,7 +1061,12 @@ function processStyle(glStyle, mapOrGroup, styleUrl, options) {
     } else {
       id = glLayer.source || getSourceIdByRef(glLayers, glLayer.ref);
       // this technique assumes gl layers will be in a particular order
-      if (!id || id != glSourceId) {
+      if (
+        // This line is because rasters set properties on their source
+        glLayer.type === 'raster' ||
```
orangemug, line 1066: This is the line I'm most unsure about. We need this because is we have two layers with the same source they need to apply different operations to their sources.

orangemug, line 1066: This is the line I'm most unsure about. We need this because is we have two layers with the same source they need to apply different operations to their sources.

orangemug:
Tests are failing due to the addition of `RasterSource`, if y'all are happy with the approach, I'll work on fixing them. Let me know.

ahocevar:
I'm sorry, but I won't be able to review this for a while. I'm terribly busy at the moment.

orangemug:
No worries @ahocevar

orangemug:
> Tests are failing due to the addition of `RasterSource`, if y'all are happy with the approach, I'll work on fixing them. Let me know.

Updated this to support more of the spec, if you are happy I'll fix up the tests.

orangemug:
> I'm failing to understand why existing tests are failing

@ahocevar I'm still doing some hunting on this one, hence still in draft. Essentially previously a raster was a 

> `TileLayer` with a `TileJSON` source

Now it's a 

> `ImageLayer` with a `Raster` source. The `Raster` source has the `sources` array set to a `TileLayer` with a `TileJSON` source.

So essentially there is this new wrapper, where `tileLayer` below is the same as it was previously.

```js
  const layer = new ImageLayer({
    source: new Raster({
      operationType: 'image',
      operation: rasterShader,
      sources: [tileLayer],
    }),
  });
```

So the _source_ in the following line of the tests now refers to the `Raster` rather than `TileJSON` source.

```js
      it('creates the correct tile grid for raster sources', function (done) {
        apply(target, context)
          .then(function (map) {
            const statesSource = map.getLayers().item(0).getSource();  // <======== HERE
            const statesTileGrid = statesSource.getTileGrid();
            should(statesTileGrid.getTileSize()).eql(256);
            should(statesTileGrid.getExtent()).eql([
              -20037508.342789244, -20037508.342789244, 20037508.342789244,
              20037508.342789244,
            ]);
```

Seems like an easy fix right? But I can't appear to find the openlayers API to get to the `tileJson` (shown below) once the layer is instantiated

```js
  const layer = new ImageLayer({
    source: new Raster({
      operationType: 'image',
      operation: rasterShader,
      sources: [tileLayer],    // <=== Not sure how to get at this from the test suite.
    }),
  });
```

I'm happy doing some more hunting, but it'll probably be a couple of days, as I'm a little busy right now.

ahocevar:
If I understand you correctly, a `raster` layer is now always an `Image` layer with a `raster` source? If so, it's not a good idea. Fine if any raster transformations that require a raster source are applied, bit without that, we should continue to use the Tile layer with the TileJSON source. Simply because the Raster source adds a significant performance penalty.

orangemug:
> If I understand you correctly, a raster layer is now always an Image layer with a raster source? If so, it's not a good idea. Fine if any raster transformations that require a raster source are applied, bit without that, we should continue to use the Tile layer with the TileJSON source. Simply because the Raster source adds a significant performance penalty.

Yeah that makes sense, I'll add a branch for raster without operations. I'll try and add some new tests for the raster operations as well.

ahocevar:
@orangemug Do you have any plans to pick this up again?

PR #1057, opened 2023-12-13
Don't throw on unsupported `heatmap` layer type instead show helpful debug info

orangemug:
Previous to this change a style with a `heatmap` layer type would "throw" crashing the app. We now just `console.debug` with

```
[ol-mapbox-style] layers[1].type "heatmap" not supported
```

Reasons for `console.debug` instead of console warn/log, is that `console.debug` doesn't show up in the console by default. You have to enable verbose logging in the web inspector.

This also has the same behaviour as other unsupported styling rules, for example [icon-pitch-alignment](https://docs.mapbox.com/style-spec/reference/layers/#layout-symbol-icon-pitch-alignment)

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -1030,9 +1030,7 @@ function processStyle(glStyle, mapOrGroup, styleUrl, options) {
     if (type == 'heatmap') {
       //FIXME Unsupported layer type
       // eslint-disable-next-line no-console
-      console.debug(
-        `[ol-mapbox-style] layers[${i}].type "${type}" not supported`
-      );
+      console.debug(`layers[${i}].type "${type}" not supported`);
       continue;
     } else {
       id = glLayer.source || getSourceIdByRef(glLayers, glLayer.ref);
--- a/src/apply.js
+++ b/src/apply.js
@@ -1029,7 +1029,11 @@ function processStyle(glStyle, mapOrGroup, styleUrl, options) {
     const type = glLayer.type;
     if (type == 'heatmap') {
       //FIXME Unsupported layer type
-      throw new Error(`${type} layers are not supported`);
+      // eslint-disable-next-line no-console
+      console.debug(
+        `[ol-mapbox-style] layers[${i}].type "${type}" not supported`
+      );
+      continue;
     } else {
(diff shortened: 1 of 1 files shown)


ahocevar:
I'd prefer `console.error` over `console.debug`. It is easier to see when something goes wrong for those who have a developer console open, and others won't bother anyway.

orangemug:
The only problem with `console.error` is it's very noisy. So if you load a style that's has a `heatmap`, it is a valid style but it'll complain a lot. That means that if a person uses this library in their map (like I do) I get spammed with errors I have no way of currently resolving. For me this is on every style change as I update styles a lot when editing.

Also it's a little odd that layer types error when unsupported but unsupported properties don't.

Other options, worth considering

 1. Just don't log anything, which is in keeping with the rest of the libraries properties (later we could add a function to report/remove the unsupported properties, once #10 is complete)
 2. Log via [debug](https://www.npmjs.com/package/debug) which isn't noisy by default, would require prod / non-prod builds (probs too complex for now)
 3. @orangemug just go ahead and use `console.error(..)`
 
Please pick an option (1, 2, or 3)

src/apply.js
```diff
@@ -1029,7 +1029,11 @@ function processStyle(glStyle, mapOrGroup, styleUrl, options) {
     const type = glLayer.type;
     if (type == 'heatmap') {
       //FIXME Unsupported layer type
-      throw new Error(`${type} layers are not supported`);
+      // eslint-disable-next-line no-console
+      console.debug(
+        `[ol-mapbox-style] layers[${i}].type "${type}" not supported`
```
ahocevar, line 1034: I don't think it is necessary to display `[ol-mapbox-style] ` - the console shows the origin of the message.

ahocevar:
For the case of things that are not supported by ol-mapbox-style, I think you're right that a debug message is appropriate.

src/apply.js
orangemug: I've made this change, resolving.

orangemug:
Hopefully this should be good to merge now.

ahocevar (approved):

PR #1060, opened 2023-12-14
Added `*-sort-key` support

orangemug:
This PR implements the `*-sort-key` rules.

Here is an example with the following layer rule, where `#ff0000` fills have a higher sort order that `#0000ff` fills. 

```json
{
  "id": "test",
  "type": "fill",
  "source": "polygons",
  "layout": {
    "fill-sort-key": [
      "get",
      "zIndex"
    ]
  },
  "paint": {
    "fill-color": [
      "get",
      "color"
    ]
  }
}
```

Previously this resulted in 

Screenshot from 2023-12-14 16-20-25

Now we get the correct result. 

Screenshot from 2023-12-14 16-21-24

### DIFF
--- a/src/apply.js
+++ b/src/apply.js
@@ -28,6 +28,7 @@ import {
   stylefunction as applyStylefunction,
   getValue,
   styleFunctionArgs,
+  types,
 } from './stylefunction.js';
 import {bbox as bboxStrategy} from 'ol/loadingstrategy.js';
 import {createXYZ} from 'ol/tilegrid.js';
@@ -50,6 +51,7 @@ import {
 } from 'ol/proj.js';
 import {getFonts} from './text.js';
 import {getTopLeft} from 'ol/extent.js';
+import {getUid} from 'ol';
 import {hillshade} from './shaders.js';
 import {
   normalizeSourceUrl,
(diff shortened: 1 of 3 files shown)


orangemug:
@ahocevar this failing test seems to be caused by a flakey test.

I can get this to pass/fail locally by just running `npm test` multiple times. Any ideas?

ahocevar:
The test issue should have been fixed a few days ago. Maybe just rebase this pull request?

orangemug:
This branch is all up to date and has the latest commit `4d4aeaa` (on `main`)

```bash
# git status | grep "On branch"
On branch feature/add-sort-key-support
# git log --oneline | grep 4d4aeaa
4d4aeaa Merge pull request #1050 from openlayers/dependabot/npm_and_yarn/html-webpack-plugin-5.5.4
# git push origin feature/add-sort-key-support
Everything up-to-date
```

ahocevar:
At first glance it looks to me like the tests fail correctly because this pull request breaks something. But I don't have time at the moment to give it a closer look.

ahocevar:
I see what's wrong. The current sort (`index`) makes sure layers as defined in the Mapbox style are rendered in the correct order (i.e. the order they appear in the `layers` array). This pull request breaks that.

If I read the `*-sort-key` Property correctly, it should sort the features so they appear in the correct order *within* the layer. To achieve that, the features should be sorted using the `renderOrder` function of the `ol/layer/Vector`. Unfortunately there is no equivalent for this `renderOrder` function for `ol/layer/VectorTile`.

orangemug:
I'll have another proper look tomorrow @ahocevar. But FYI I still see flakiness with tests. I get passes/fails of the test suite on repeat runs without any changes

```bash
# npm test
Chrome Headless 119.0.6045.105 (Linux x86_64): Executed 128 of 176 SUCCESS (2.07 secs / 1.928 secs)
TOTAL: 128 SUCCESS

=============================== Coverage summary ===============================
Statements   : 62.41% ( 890/1426 )
Branches     : 62.91% ( 570/906 )
Functions    : 74.11% ( 126/170 )
Lines        : 63.09% ( 865/1371 )
# echo $?
0
```

orangemug:
I think within the layer is working correctly, at least it should be will confirm tomorrow. The function uses a secondary index multiplied by a tiny number.

```js
export function calcSortIndex(index, sortIndex) {
  // The `sortIndex` here is the index within the layer, we multiply that by
  // a tiny number so we end up with a small offset within the current layers
  // bounds
  return index + (sortIndex === undefined ? 0 : sortIndex) * 0.00000000001;
}
```

ahocevar:
Using a different zIndex for every feature adds a significant performance penalty. Using the layer's `renderOrder` function would be a better choice.

orangemug:
> Using a different zIndex for every feature adds a significant performance penalty. Using the layer's renderOrder function would be a better choice.

Ok cool, will change over tomorrow.

Regarding tests @ahocevar. I'm seeing some sort of odd object mutation thing (**I think**) this line seems to differ on some runs of the test suite, `propertySpec` is sometimes `undefined`. I'll see if I can work what's going on.

https://github.com/openlayers/ol-mapbox-style/blob/bd5fe28b82071ad1da8e5996f5bd491c2aa1156b/src/stylefunction.js#L120

ahocevar:
I'm currently reworking decluttering and z-index ordering in OpenLayers. Let's revisit when that is done.

ahocevar:
@orangemug Would you like to take another look at this, now that z-index is handled way smarter in OpenLayers?

PR #1074, opened 2024-01-11
Zero text halo should be no halo

toddwong:
Zero width text halo should be treated as no text halo


ahocevar (approved):
Thanks, @toddwong

PR #1169, opened 2024-06-28
Remove polyfill io

orangemug:
Because of security issues outlined in <https://dev.to/snyk/polyfill-supply-chain-attack-embeds-malware-in-javascript-cdn-assets-55d6>

Removed because no longer required, browser support is good enough now


ahocevar (approved):
Thanks, @orangemug

PR #1205, opened 2024-10-15
more precise angle computation

leyan:
Currently, when using `"symbol-placement": "line"` and `"icon-rotation-alignment": "map"`, the angle of the icon is computed in ol-mapbox-style, as we know the location of the icon placement but not the corresponding line segment information. So the computation first tries to determine on which line segment the icon is located, then computes the slope of this segment and uses that as the rotation of the icon.

However, the algorithm used to determine on which segment is the icon is too simplistic: it simply takes the first segment for which the icon is inside the bounding box. However, there is no guarantee that this segment is the correct one and it is quite easy to construct a counter-example giving a wrong rotation angle: 
image 

while it should be 

image

This pull request proposes to replace this bounding box by a more precise comparison. I used to 0.001 tolerance to avoid dealing with potential floating point unpleasantness, I don't know if it is the ideal value but it is anyway much more precise than the current version.

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -832,14 +832,15 @@ export function stylefunction(
                         const x2 = coordinates[i + stride];
                         const y2 = coordinates[i + stride + 1];
                         const minX = Math.min(x1, x2);
-                        const minY = Math.min(y1, y2);
                         const maxX = Math.max(x1, x2);
-                        const maxY = Math.max(y1, y2);
                         const xM = midpoint[0];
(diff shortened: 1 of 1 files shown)


ahocevar (changes requested):
Going forward, I think a `placement: line` option for `ol/style/Image` would make sense, to avoid the potentially expensive calculation here. But for now, your code makes sense. Please fix the linter errors though.

ahocevar (approved):
Thanks, @leyan

PR #1206, opened 2024-10-15
Manage text-keep-upright:false

leyan:
Currently, the `text-keep-upright` setting is ignored, as OpenLayers automatically keeps the labels upright and does not expose an option to control it.

This pull request offers to manually compute a rotation angle for the label when `text-keep-upright:false`. 
It means that we have to force `"symbol-placement":"point"`, so a single label is written on the middle point of the line instead of several for the better `line` placement but there should be no risk of regression as at the moment this specific option is not managed at all.

I reused the same algorithm currently used to rotate icons, maybe we should avoid duplication and write a common function?

Use case: I am trying to use arrow symbols to represent the direction of a line and I would prefer to use labels rather than icons for that. This can only work with `text-keep-upright:false`.

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -1388,19 +1388,6 @@ export function stylefunction(
               : wrapText(label, font, maxTextWidth, letterSpacing);
           text.setText(wrappedLabel);
           text.setFont(font);
-          text.setRotation(
-            deg2rad(
-              getValue(
-                layer,
-                'layout',
-                'text-rotate',
-                zoom,
-                f,
-                functionCache,
-                featureState,
-              ),
-            ),
-          );
           const textAnchor = getValue(
             layer,
             'layout',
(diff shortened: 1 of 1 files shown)


ahocevar (changes requested):
@leyan Thanks for your contribution. ~~Instead of your custom code, please set `rotateWithView: true` option on `ol/style/Text` when `text-keep-upright: false` is set. This will be a much simpler change then.~~

**Update**: sorry, I thought `text-rotation-alignment` instead of `text-keep-upright`. Anyway, in OpenLayers this change would be much simpler, so I'd encourage you to submit a pull request there, adding an `avoidUpsideDown` option to `ol/style/Text`. I'll post another comment shortly pointing you to the only line in the code that needs to be changed there.

ahocevar:
To introduce a `keepTextUpright` option on `ol/style/Text` in OpenLayers, it would need to be passed to `ol/geom/flat/textpath.js`'s `drawTextOnPath()` function. Inside that, `reverse` would always need to be `false` when `keepTextUpright` is set to `false`:

https://github.com/openlayers/openlayers/blob/9e574dac382a376c3fb42fd80fc4268c8dcd3f63/src/ol/geom/flat/textpath.js#L72-L80

leyan:
Thank you for the feedback. I agree that it is better to offer an option in OpenLayers itself but I was afraid that the change would be harder to pass. I will try to do something if you think it is the best way.

PR #1208, opened 2024-10-20
Use the new keepUpright parameter in OpenLayers

leyan:
Until now, the `text-keep-upright` parameter in the Mapbox Style specification was ignored, as OpenLayers was automatically keeping the labels upright and did not expose an option to control it. Now that the `keepUpright` parameter is available on `ol/style/Text`, this PR uses it to cover `text-keep-upright`.

This is a follow-up to #1206 now that the cleaner solution is available.

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -1402,6 +1402,16 @@ export function stylefunction(
               ),
             ),
           );
+          const keepUpright = getValue(
+            layer,
+            'layout',
+            'text-keep-upright',
+            zoom,
+            f,
+            functionCache,
+            featureState,
+          );
+          text.setKeepUpright(keepUpright);
           const textAnchor = getValue(
             layer,
             'layout',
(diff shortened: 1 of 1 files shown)


src/stylefunction.js
```diff
@@ -1402,6 +1402,16 @@ export function stylefunction(
...
+          const keepUpright = getValue(
+            layer,
+            'layout',
+            'text-keep-upright',
+            zoom,
+            f,
+            functionCache,
+            featureState,
+          );
+          text.setKeepUpright(keepUpright);
```
ahocevar, line 1414: For compatibility with older OpenLayers versions, this should be wrapped in an if-block:
```js
if (typeof text.setKeepUpright === 'function') {
  // ...
}

ahocevar (changes requested):

leyan:
@ahocevar Should I create a test for this parameter? if so, how should I go about it? I saw there are quite a lot of tests but I am not sure about how to add one.

ahocevar (approved):
Thanks, @leyan

PR #1220, opened 2024-11-25
Update stylefunction.js to fix textField reduce function

### LINKED ISSUE #1217: acc is undefined in stylefunction text-field
I'm using this library to create OL-compatible styles from `protomaps-themes-base` (https://github.com/protomaps/basemaps).  I apply the style to a layer with a `PMTilesVectorSource` source using:
```js
stylefunction(layer, {
	version: "8",
	layers: protomaps_themes_base("protomaps", "light"),
	sources: { protomaps: { type: "vector" }}
}, 'protomaps')
```
At certain zoom levels, the style is referencing some text labels, and that triggers the first part of the bug here: https://github.com/openlayers/ol-mapbox-style/blob/main/src/stylefunction.js#L1298

The function passed to `textField.sections.reduce`  should `return acc` on that line instead of just `return`.  The next iteration of the function receives the previous accumulator value of `acc`, which is now `undefined`, so things like `acc.push()` will throw an error.

There's more to my issue than that, but I'm still trying to determine whether it's related to the style itself.

mzmcgrat:
Fixes a bug to have the textField reduce function return the accumulator in all cases. Fixes #1217

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -1295,7 +1295,7 @@ export function stylefunction(
                 }
                 if (type == 2) {
                   acc.push(applyLetterSpacing(text, letterSpacing), chunkFont);
-                  return;
+                  return acc;
                 }
                 text = wrapText(
                   text,


ahocevar (approved):
Thanks, @mzmcgrat

PR #1228, opened 2024-12-12
Get rid of expandUrl dependency

### LINKED ISSUE #1227: expandUrl is not a function error
I'm trying to use the applyStyle function to a VectorTile layer and it returns an error:
util.js:242 Uncaught (in promise) TypeError: f.expandUrl is not a function
I'm using javascript version of OpenLayers  from [jsdelivr.net cdn](https://cdn.jsdelivr.net/npm/ol/dist/ol.js)

image

This is the way i've implemented it:

image
image

ahocevar:
Fixes #1227.


ahocevar:
The second commit is unrelated, but necessary to make the tests pass again on GitHub.

PR #1260, opened 2025-02-21
quick fix for openlayers/ol-mapbox-style#1233

prusswan:
This suppresses the non-printable labels, e.g. 

```
[ "", 'normal 400 12px/1.2 "Noto Sans"', "\n", "", "", 'normal 400 12px/1.2 "Noto Sans"', "\n", "", "", 'normal 400 12px/1.2 "Noto Sans"' ]
```

### DIFF
--- a/src/stylefunction.js
+++ b/src/stylefunction.js
@@ -1321,6 +1321,10 @@ export function stylefunction(
                 return acc;
               }, []);
             }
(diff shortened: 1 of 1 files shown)


ahocevar:
Thanks for your effort @prusswan, in the meantime I created https://github.com/openlayers/openlayers/pull/16665 to make this error go away in OpenLayers, and convert the rich text style to a plain text style.

PR #1267, opened 2025-03-03
Fix vector sources lookup

remogeissbuehler:
## Overview
This PR introduces a heuristic fix to choose the "correct" source layer for vector maps. In particular, when the `style.json` has multiple sources, such as:

```json
  "sources": {
    "attribution_only": {
      "attribution": "<a href=\"https://example.com/\" target=\"_blank\">&copy; Example</a> ",
      "type": "vector"
    },
    "world": {
      "url": "https://example.com/tiles.json",
      "type": "vector"
    }
  },
```

this currently fails, as the logic to pick the source to display for a `VectorTileLayer` is "take the first source with `type == 'vector'`":

https://github.com/openlayers/ol-mapbox-style/blob/7b69c36a9792581b5d479c24ad9a2789a91593c3/src/apply.js#L253C1-L255C14

So in this case, we pick the one that does not in fact have any tile reference (`.url` or `.tiles`), which in turn leads to an exception in `utils.js`:
Screenshot 2025-03-03 at 01 29 03

## Motivation
This exact bug happens for the Maptiler "basic-v2" map. I included it as a new example, which is broken in the current version but works with this fix.

## Implementation
To tackle this problem, I added two main fixes:

### Check for tiles and url in utils (`getTileJson`)
`getTileJson` previously assumed that either `.url` or `.tiles` are present. Sadly, this assumption doesn't always hold, so I added an extra check and raise an Error when neither are present for a more meaningful message in case this happens.

### Heuristic choice of source
I replaced the above-mentioned logic to choose the source for `VectorTileLayer` for the following: "choose the first source of correct type that has at least one of the required attributes":
https://github.com/remogeissbuehler/ol-mapbox-style/blob/dc21b6e2783451fa5f3190c38b95a60053f4a6b2/src/apply.js#L259C1-L269C14

Note: I don't think this fully fixes the issue, as there could be multiple relevant sources and we still only pick one (at least a relevant one now). If I understand [this spec](https://maplibre.org/maplibre-style-spec/sources/#sources) correctly, the sources on their own don't say much about their display and the layers should instead be used as they refer to the different sources. Implementing this seems like a much bigger change to me though...


src/apply.js
```diff
@@ -248,10 +248,24 @@ export function applyStyle(
...
-            return glStyle.sources[key].type === type;
+            // NOTE: this is mostly a heuristic way of picking the source, the TODO comment above still applies.
+            const sourceType = glStyle.sources[key].type;
+            const typeMatch = sourceType === desiredType;
+            const hasRequiredAttrs = requiredAttrs[sourceType].some(
+              function (attr) {
+                return Object.keys(glStyle.sources[key]).includes(attr);
+              },
+            );
+            return typeMatch && hasRequiredAttrs;
```
ahocevar, line 268: Instead of this involved change, it would be smarter to detect the source by the layers that use it, i.e.
```js
          sourceId = glStyle.layers.find(function (layer) {
            return layer.source && glStyle.sources[layer.source].type === type;
          }).source;
```

examples/openmaptiles-v2-layer.js
ahocevar, line 1: Please change the existing example to use the basic-v2 style, instead of creating another example here.

ahocevar (changes requested):

examples/openmaptiles-v2-layer.js
remogeissbuehler: done in 2a4f65bbeffcdc3dc97379bf34bebee04642f430, updated the non-layer openmaptiler example as well.

src/apply.js
remogeissbuehler: I agree that this is more elegant, thanks!

remogeissbuehler:
Thanks a lot for the super quick review :) Sorry, I was quite busy this week. 

I implemented your suggestions and also updated the test cases and documentation string to reflect them. Thanks also for pointing out that the source can be configured, I had missed that 🤦🏼

src/util.js
```diff
@@ -301,6 +301,8 @@ export function getTileJson(glSource, styleUrl, options = {}) {
         tileJson: Object.assign({}, glSource),
         tileLoadFunction,
       });
+    } else {
+      promise = Promise.reject(new Error('source has no tiles'));
```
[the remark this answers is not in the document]
remogeissbuehler: I was just thinking: should we add a hint like "consider choosing a different source" or something for the next person that misses it like I did? :eyes:

src/apply.js
```diff
@@ -173,7 +173,7 @@ function completeOptions(styleUrl, options) {
  * `source` key or an array of layer `id`s from the Mapbox/MapLibre Style object. When a `source` key is
  * provided, all layers for the specified source will be included in the style function. When layer
  * `id`s are provided, they must be from layers that use the same source. When not provided or a falsey
- * value, all layers using the first source specified in the glStyle will be rendered.
+ * value, the source is inferred from the first layer.
```
ahocevar, line 176:
```suggestion
 * value, all layers using the same source as the first layer matching the provided `layer` type will be rendered.
```

src/apply.js
```diff
@@ -250,9 +250,10 @@ export function applyStyle(
 
         const type = layer instanceof VectorTileLayer ? 'vector' : 'geojson';
         if (!sourceOrLayers) {
-          sourceId = Object.keys(glStyle.sources).find(function (key) {
-            return glStyle.sources[key].type === type;
-          });
+          sourceId = glStyle.layers.find(function (layer) {
+            return layer.source && glStyle.sources[layer.source].type === type;
+          }).source;
+
```
ahocevar, line 256:
```suggestion
if (!sourceId) {
  return reject(new Error(`GL Style has no layers using source type "${type}"`));
}
```

src/util.js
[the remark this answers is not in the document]
ahocevar: You won't get here any more in this case.

ahocevar (changes requested):

src/apply.js
remogeissbuehler: is that really necessary? just a few lines down:
https://github.com/openlayers/ol-mapbox-style/blob/3c29d7ab3c94d147b5f267ff845b826073436ccb/src/apply.js#L265-L267

we have this check anyway, no?

ahocevar (approved):
Thanks, @remogeissbuehler

PR #1365, opened 2025-10-22
implement basic support for usage in worker

RobertOrthofer:
Basic support for usage in a worker. 
Because of the nature of the worker environment, this does not support css font handling.


ahocevar (approved):
Nice improvement, thanks @RobertOrthofer!

Issue #1383, opened 2025-12-07
Add support for GeometryCollection feature type


shaietz:
Currently, features of type GeometryCollection are not supported, as seen in stylefunction.js types declaration:
`const types = {
  'Point': 1,
  'MultiPoint': 1,
  'LineString': 2,
  'MultiLineString': 2,
  'Polygon': 3,
  'MultiPolygon': 3,
};`

when a feature of type GeometryCollection is encountered, ` const type = types[featureGeometry.getType()];` (stylefunction.js line 556) is undefined and no styles are applied.

adding support for GeometryCollection is a matter of iterating over the inner geometries and the styles based on the inner geometry type.

shaietz:
Yes my source is geojson
Ill start working on a PR

shaietz:
i opened a PR: #1384

PR #1384, opened 2025-12-09
feat: added support for GeometryCollections

shaietz:
Issue #1383 add support for GeometryCollection Feature type

modified stylefunction to handle GeometryCollections by calling the stylefunction on each geometry individually

### DIFF
--- a/examples/data/geojson-inline.json
+++ b/examples/data/geojson-inline.json
@@ -62,20 +62,23 @@
         "features": [ {
           "type": "Feature",
           "properties": {
-            "n": 2,
+            "n": 1,
             "cat": 1
           },
-          "geometries": [{
+          "geometry": {
             "type": "GeometryCollection",
-            "coordinates": [0.6, 0.6]
-          },{
-            "type": "GeometryCollection",
-            "coordinates": [0.7, 0.7]
-          },{
-            "type": "GeometryCollection",
-            "coordinates": [0.8, 0.8]
-          }]
-        },{
+            "geometries": [{
+              "type": "Point",
+              "coordinates": [0.5, 0.5]
+            }, {
+              "type": "Point",
+              "coordinates": [1.5, 1.5]
+            }, {
+              "type": "Point",
+              "coordinates": [2.5, 2.5]
+            }]
+          }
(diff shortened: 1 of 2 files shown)


examples/data/geojson-inline.json
```diff
@@ -59,7 +59,23 @@
...
+            "type": "GeometryCollection",
+            "coordinates": [0.6, 0.6]
+          },{
+            "type": "GeometryCollection",
+            "coordinates": [0.7, 0.7]
+          },{
+            "type": "GeometryCollection",
+            "coordinates": [0.8, 0.8]
+          }]
+        },{
```
ahocevar: These are not valid geometryCollection geometries.

src/stylefunction.js
```diff
@@ -554,7 +556,28 @@ export function stylefunction(
...
+
+      for (const geom of geoms) {
+        const subFeature = new Feature({
+          ...feature.getProperties(),
+          geometry: geom,
+        });
+
+        const subStyles = styleFunction(subFeature, resolution, onlyLayer);
+
+        if (subStyles && Array.isArray(subStyles)) {
```
ahocevar: No `Array.isArray()` check necessary here. `styleFunction()` returns `undefined` or an array of styles.

ahocevar (changes requested):
Thanks for your work on this, @shaietz. The starting point would be to create a geojson with valid GeometryCollections, preferably one with two simple geometries and one with a nested geometry collection and a simple geometry.

A test would also be crucial to make problems with the added code obvious.

I added comments for each of the problems I see from looking at the code. Please address them. Let me know when you have questions.

src/stylefunction.js
```diff
@@ -554,7 +556,28 @@ export function stylefunction(
...
+          styles.push(...subStyles);
+        }
+      }
+
+      return styles.length > 0 ? styles : undefined;
```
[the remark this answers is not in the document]
shaietz: i'm not sure i understand what you mean, anyways i have removed the push of the substyle to the styles array because it was not necessary, so is this still relevant?

